PR #4, opened 2013-05-15
items.js

yangit:
Sorry Tyler.
Items are js file now. I still can't figure out how to include coffee as submodules from within node.js so it does not conflict with require.js

Bonus: we do not have to pass items to scoring function and now there is only one file to require.


lefnire:
How were you able to get algos & helpers working as coffee, but not items? I'm going to be working today and tomorrow pretty heavily on consolidating our efforts & migrating the web code changes into shared
(sent that message before, but it's not showin up for some reason)

yangit:
Huh...
Better we chat about it.
In short, you include .coffee script to node.js by compiling it with grunt
first (you basically include js file)
I include it to browser using "cs" plug in for require JS, so grunt is not
needed.

Then we need to include another coffee script from within this file.
And syntax for that would be different for node.js and require.js

We could use coffeescript plugin for node.js or add IF to see wether we are
in node or in browser to conditionally require coffee scripts in different
ways.
But I decided not to overcomplicate this.

We still can do it later though.

PR #6, opened 2013-05-17
Coffeescript all over!

yangit:
@lefnire you gotta like it :)
See examples.
I did not convert any files under /script/ to CS since you did not commit anything last night.


lefnire:
fantastic!

PR #9, opened 2013-05-29
Check flags.rest

dendrobatid:
line 167 - Gets the flag to determine if user is resting
line 213 - if the user is resting, dailies have no effect

https://github.com/lefnire/habitrpg/issues/997


lefnire:
let me check into this, i have a hunch something else is going on. hang tight

lefnire:
yea, it's something different. if you look at https://github.com/HabitRPG/habitrpg-shared/blob/master/script/algos.coffee#L336, we bail out of running cron altogether (which triggers the scoring functions) if they're resting in the inn. Have you got it running locally? If so, could you per chance dig around that area? see if that section's getting skipped for some reason? I'm gonna close this PR because score() should never be called on cron when resting anyway.

PR #16, opened 2013-09-03
Optimizations

paglias:
This pull request do:
- remove an enormeous data-uri source map for the coffee files which was taking more than 400kb of data. Now the file from more than 700kb is only 290kb
- minify css here on habitrpg shared because minification in other places (ie the mobile build) would end in wrong paths for all the sprite images
- js instead can be minifed in normal builds (ie mobile build)
- remove a duplicate spritesheets.css from root and use only the /dist version which is created with Grunt so no need to update manually

Now, after this changes css paths need to be updated. I pushed the necessary changes at HabitRPG/habitrpg-mobile#120 and lefnire/habitrpg#1413


lefnire:
this is great @paglias , thanks!!

PR #21, opened 2013-09-24
Corrected rounding of gold & silver calculations

djuretic:
This should fix https://github.com/lefnire/habitrpg/issues/1560.
For example, if I have 1.957 "coins", under the old code that is converted to 2 gold and 96 silver, instead of 1 gold and 95 silver.

### DIFF
--- a/script/helpers.coffee
+++ b/script/helpers.coffee
@@ -269,7 +269,7 @@ module.exports =
   ###
   silver: (num) ->
     if num
-      Math.floor (num - Math.floor(num))*100
+      ("0" + Math.floor (num - Math.floor(num))*100).slice -2
     else
       return "00"
 
--- a/script/helpers.coffee
+++ b/script/helpers.coffee
@@ -260,7 +260,7 @@ module.exports =
   ###
   gold: (num) ->
     if num
-      return (num).toFixed(1).split('.')[0]
+      return Math.floor num
     else
       return "0"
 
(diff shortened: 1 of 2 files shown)


script/helpers.coffee
```diff
@@ -269,7 +269,7 @@ module.exports =
   ###
   silver: (num) ->
     if num
-      (num).toFixed(2).split('.')[1]
+      ("0" + Math.floor (num - Math.floor(num))*100).slice -2
```
lefnire, line 272: oh, that's clever - the "0" with the slice(-2). PS `~~num`===`Math.floor(num)` if you want some more trickery.

I'm not following this `Math.floor (num - Math.floor(num))*100` bit though. Why doesn't `num.toFixed(2)` work, and what does your change do instead?

djuretic, line 272: Because sometimes `toFixed` rounds up:

```
>>> (1.974).toFixed(2)
"1.97"
>>> (1.974).toFixed(2).split('.')[1]
"97"

>>> (1.975).toFixed(2)
"1.98"
>>> (1.975).toFixed(2).split('.')[1]
"98"
```

In the last case, having 1.975 gp will show that you have 98 silver coins, not 97.

So I'm using Math.floor to round consistently:

```
>>> Math.floor((1.974 - Math.floor(1.974))*100)
97
>>> Math.floor((1.975 - Math.floor(1.975))*100)
97
```

PR #22, opened 2013-09-25
Adds new skin tones

Shaners:
Adds 3 new skin tones options for the 2 different body types.

Not entirely clear if female_sprites.css and male_sprites.css are used at all because they duplicate what is in spritesheets.css. Maybe spritesheets.css is calculated from the other two?


djuretic:
It appears that there is a grunt task that generates the file spritesheets.css: https://github.com/HabitRPG/habitrpg-shared/blob/rewrite/Gruntfile.js#L4

Try running `grunt cssmin` from the command line.

paglias:
Yes, before sending request you should run `grunt` (not only cssmin so it works with changes to js files, too). But it seems it has been run, no @Shaners ?

Shaners:
Yep it has been run. Funny thing is, in my ignorance by my own fault, I manually made the same change to spritesheets.css that was probably over-written by running `grunt`. I didn't quite understand why my image and jade edits would show up right away in local but style and css edits wouldn't, even with a rs. Got a better understanding now though.

For this type of PR is it better to <em>not</em> commit changes to things like dist/spritesheets.css, since that will be recalculated when grunt is ran on the server?

paglias:
Sadly we have to commit files built by grunt because habitrpg-shared is
used by the mobile app via bower, not as a node package that can be built
when running `npm install`
Il giorno 25/set/2013 18:35, "Shane Lister" notifications@github.com ha
scritto:

lefnire:
oh, this PR should be fine actually @Shaners - in fact, we can just merge this - it's just CSS additions, and it all looks great to me. It's only the website onduhray changesets that got effected. Feel free to re-open & merge

PR #25, opened 2013-10-06
Added drop system tests

djuretic:
I've added some tests concerning the drop system, I had to stub Math.random to make the tests deterministic.


lefnire:
beautiful!

PR #26, opened 2013-10-16
Adding custom backer items from Leslie.

Shaners:
Sprite sheet for custom backer gear, created by Leslie.

Not sure what the mechanics are going to be to give the backers access to these items so was unsure where to put the CSS positioning stuff.


lefnire:
shnazzy! so only 4 of them pulled through eh?

Shaners:
I think only 3 responded, the rest will be getting that helmet. Maybe seeing a new random piece of gear no one else has will prompt them to respond. Or can use them as a playground to test out new gear ideas.

lefnire:
oh, smart with the "fallback item". Ok, I'll start brainstorming an implementation strategy.

Shaners:
Perhaps perma-equiped with custom stat bonuses (so not available in the shop) and in the future an ability to reskin but keep elevated stats. So they always have this item equipped no matter what with the highest stats plus a bonus. It'd power level them at lower levels but that's kind of the point of the gear me thinks.

-Shane

PR #30, opened 2013-10-22
Reset lastDrop to zero when day ends

paglias:
Discussed in https://github.com/HabitRPG/habitrpg/issues/1625#issuecomment-26829372, it resets the lastDrop's count to zero when day ends.

There are a few thing I'm not sure about this that should be checked before merging:
- user.item.lastDrop.count is always defined? I think so because it's defined in User's schema so it doesn't check for user.items.lastDrop.count to be defined
- `paths['items.lastDrop'] = true` is needed for Mongoose to see what has changed? It's ok to use `paths['items.lastDrop'] = true` even though `items.lastDrop.count` only is modified?

### DIFF
--- a/script/algos.coffee
+++ b/script/algos.coffee
@@ -371,6 +371,11 @@ obj.cron = (user, options={}) ->
 
   user.lastCron = now; paths['lastCron'] = true
 
+  # Reset the lastDrop count to zero
+  if user.items.lastDrop.count > 0
+    user.items.lastDrop.count = 0
+    paths['items.lastDrop'] = true
+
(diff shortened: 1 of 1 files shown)


lefnire:
Yeah, so anything defined in the Mongoose schema will be provided with defaults when the user is loaded, allowing us to avoid many errors (one of the reasons I was so keen on getting User.tasks converted to User.habits/dailys/etc - but that's for later). The paths thing is unnecessary as per our conversation (which happened after this thread, sorry about that)

lefnire:
tested around with this, this looks great. I'm surprised it was working on the old system :o I'm going to push this to develop & master

PR #31, opened 2013-10-24
New Red Hair Option

Shaners:
Also fixed a few pixel grid alignment issues.


lefnire:
i'll push to dev. may be a bit before pushing to master, since we'll want to cut a new mobile release too & we'll probably want a handful more changes in mobile before pushing

Shaners:
Sounds good! Beta web will get this soon though? Edge users need their edge features!

lefnire:
pushing to beta now, 5m

lefnire:
on prod too, but you can't select the hair. eh, maybe we should just roll with it if we don't have a reason to cut a new mobile after a few days

Shaners:
Just noticed that the female red hair is missing pixels. This will need to be fixed so maybe avoid pushing to mobile until then.

PR #34, opened 2013-10-27
Fixed unit tests

djuretic:
I've fixed the test so the Travis build is back to normal.
I've noticed that the `startOfWeek` helper depends on the locale, if any future feature is going to depend on it, we should revisit it.

### DIFF
--- a/.travis.yml
+++ b/.travis.yml
@@ -1,6 +1,3 @@
 language: node_js
 node_js:
   - "0.10"
-branches:
-  only:
-    - rewrite
--- a/README.md
+++ b/README.md
@@ -20,6 +20,9 @@ Shared resources useful for the multiple HabitRPG repositories, that way all the
   * `algos.score(user, task, direction)`, etc
   * TODO document all the functions
 
+##Tests
+* `npm install`
+
 ##CSS
 Shared CSS between the website and the mobile app is a fuzzy area. Spritesheets definitely go in habitrpg-shared (since mobile
 uses them too). Other things, like customizer buttons, *may* want to go here? As you find sharable components, (1) move them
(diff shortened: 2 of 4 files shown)


djuretic:
I've also added how to run tests (`npm test`) in the readme file.

djuretic:
Should I create a new pull request targeting the master branch instead of the rewrite branch?

lefnire:
yeah, do this one against master, that's deployed now on beta/production & we'll kill rewrite soon. I added you as a repo admin (shared & web) so I won't be a bottleneck to any contributions you think don't need review per se, so feel free to merge to master as well

djuretic:
Great! 
The only thing preventing me for merging this is here: https://github.com/djuretic/habitrpg-shared/blob/8743e00b623fc96f0fbe3fe2adf63ae301969c1b/script/helpers.coffee#L79
I've added `lastDrop: { date: +new Date, count: 0 }`, is that change safe? I don't know how widely the site relies on `helpers.newUser` defaults.

lefnire:
Looks legit to me! The `helpers.newUser` function is indeed used, kinda inconsistently though. Since moving to Mongoose (from Derby), we're able to rely on its default values for new documents (see https://github.com/HabitRPG/habitrpg/blob/develop/src/models/user.js#L138). However, I don't think we've fully made the switch, so newUser is used from time to time. It's also handy to have around if we ever want to revisit the feature where users can play with the site before registration (they used to be able to, then we had to remove due to maintenance issues). 

Anyway, all looks good. I'd say do another PR to master

lefnire:
i merged to master actually, gonna jump into some tests stuff

PR #40, opened 2013-11-15
Fix drop ratios

therufs:
Ratios left intact as commented in algos.coffee.  

But this means that potion drop odds are as follows:  

Gold:  10% 
CC Pink, CC Blue, Zombie: 6.66% each (6.66 \* 3 = 20% total chance)
Red, Shade, Skeleton:  10% each ( \* 3 = 30%) 
Base, White, Desert:  13.33% each ( \* 3 = 40%)

Suggest (with mixed feelings) reducing Very Rare to 4% and distributing the other 6% between other three tiers.  Will write & submit that patch too; pick your -poison- potion!

### DIFF
--- a/dist/habitrpg-shared.js
+++ b/dist/habitrpg-shared.js
@@ -8336,7 +8336,7 @@ if(typeof module != 'undefined' && module.exports){
         drop.type = 'Egg';
         drop.dialog = "You've found a " + drop.text + " Egg! " + drop.notes;
       } else {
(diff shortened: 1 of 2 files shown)


lefnire:
looks like you went your suggestion, looks good!

lefnire:
@therufs does [the latest update](https://github.com/HabitRPG/habitrpg-shared/blob/master/script/algos.coffee#L135tL170) (adding in foods) still look ok?

therufs:
um, ... almost!  :)

Golden = 0.03 is good.  But then to make the next tier of potions actually
happen 0.06 of the time, they have to occur 0.03 < potion <= 0.09.
Likewise, for the next tier to happen 0.09 percent of the time, they should
occur 0.09 < potion <= 0.18.

Make sense?  I could probably draw a picture if it'd be helpful ...

lefnire:
Thanks friend :) Math has never been my strong suite...

PR #41, opened 2013-11-16
Use egg.text instead of egg.name & fix unit tests

djuretic:
It seems that merging the mounts branch to master broke the build, I've fixed the tests and found that egg.text should be used instead of egg.name in algos.coffee.

### DIFF
--- a/script/algos.coffee
+++ b/script/algos.coffee
@@ -143,8 +143,8 @@ randomDrop = (user, delta, priority, streak = 0, options={}) ->
     # Eggs: 30% chance
     if rarity > .5
       drop = helpers.randomVal eggs
-      user.items.eggs[drop.name] ?= 0
-      user.items.eggs[drop.name]++
+      user.items.eggs[drop.text] ?= 0
+      user.items.eggs[drop.text]++
       drop.type = 'Egg'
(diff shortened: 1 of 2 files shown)


lefnire:
Uhoh, we're gonna wanna fiddle with the tests to get this passing, instead of algos.coffee, because `egg.name` is the unique key (should never change on the backend, and should not include spaces) that's used on the `items.eggs` object, so that we have something like: 

``` javascript
user.items.eggs = {
  "Dragon-Red": 5,
  "Wolf-Base": 5
  ...
}
```

and allows us to perform clever lookup and logic throughout the app. `egg.text` on the other hand looks like "Panda Cub", and is subject to change (has in the past), and change will cause issues for the user. So we definitely want to use `egg.name` in this case, not `egg.text`.

BTW @djuretic, I apologize that I haven't been keeping up with the tests like I'd promised. You're doing a fantastic job keeping these up-to-date, and it's so god-damn essential - your commits are _extremely_ valuable. I hope to eventually get to the point that I'm keeping up with these tests with you, let me know if it ever gets to where it's frustrating you and I'll switch gears.

djuretic:
Got it, I'll revert the changes to algos.coffee. Since only the test code changed, I'll do the merge.

lefnire:
awesome, you rock @djuretic!

PR #42, opened 2013-11-17
Change "HP" to "Health" and clarify Potion type

SabreCat:
Continuing standardization of terminology, using "Health" wherever possible instead of "HP".

Also retitled the "Potion" in the basic item shop to be "Health Potion", to distinguish from Hatching Potions, and fixed a couple of mount system typos I spotted along the way.

### DIFF
--- a/script/items.coffee
+++ b/script/items.coffee
@@ -14,35 +14,35 @@ items = module.exports.items =
   ]
   armor: [
     {index: 0, text: "Cloth Armor", classes: 'armor_0', notes:'Training armor.', defense: 0, value:0}
-    {index: 1, text: "Leather Armor", classes: 'armor_1', notes:'Decreases HP loss by 4%.', defense: 4, value:30}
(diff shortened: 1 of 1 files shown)


lefnire:
same question as the https://github.com/HabitRPG/habitrpg/pull/1847 . Maybe a drupal poll? like

Do we want to 
(a) use common human terms for RPG elements, to appeal to people who don't game
(b) use game terms, to keep things in character

nafoster:
It might not be a bad idea to make a poll. I know I recently invited friends who are non-gamers that weren't sure what GP meant, but that could be an isolated case.

SabreCat:
I've seen a few Tavern questions about what "GP" means, as well. Simply "Gold" is more intuitive to non-native English speakers, too. I don't get what you mean by game terms to keep things in character, @lefnire -- plenty of RPGs use simply "Health" and "Gold" (or gil, zenny, etc.) rather than abbreviations. Would a "real" Habiteer warrior be somehow more likely to talk in initialisms than in plain language..?

PR #43, opened 2013-11-23
Updated Abilities

DanielTheBard:
Not sure how balanced they are. Will probably require some play-testing. I'm sure we'll make adjustments as we go along.
Also, built this on the assumption of maybe starting with maybe 30mp, and gaining 2 mp per point of INT?
Took a crack at conveying what I had in mind fake-code-wise, as well. Not sure how balanced that will be, but we'll see I suppose!

### DIFF
--- a/script/items.coffee
+++ b/script/items.coffee
@@ -68,107 +68,126 @@ items = module.exports.items =
   spells:
     wizard:
       fireball:
-        text: 'Fireball'
+        text: 'Burst of Flames'
         mana: 10
         target: 'task'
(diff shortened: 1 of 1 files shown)


lefnire:
perfect! Thanks for having the confidence to write the pseudo-code - I'll fix things up, and maybe throw pointers at you periodically so we can team up on refining these math-wise over time. We may need a math-guru to help us refine the balance of everything... (Hmm, I wonder if @MrConcept still plays habit..)

PR #44, opened 2013-11-27
adding emoji images

snicker:
Available for merge if this is the decided path for https://github.com/HabitRPG/habitrpg/issues/1935


lefnire:
Might be easier to `bower install https://github.com/github/gemoji#master --save` and reference `/images/emoji/unicode`. This way less clutter in the repo, plus we get updated / added emojis. Or is the css expecting /img directory, and that would break things?

Question: how is this on size? Our mobile app's size restriction is 10mb, since we're using PhoneGap Build (we can opt not to use PGB if we need, but I think ~10mb is a good soft limit anyway for now). TMK the current size is ~7mb. Can't imagine this adds 3mb, but question just in case. Anyway, we can always exclude from mobile dist build but keep on web

lefnire:
Oh, I'm sorry - I just read your bit on bower & symlinks. Yeah, this PR looks good to me then

lefnire:
alas, 12mb. We're just gonna wanna make sure these don't get copied in mobile build

snicker:
Well, they shouldn't have to, as mobile users have emoji built in ;)

lefnire:
perfect

snicker:
The other alternative of course is to just commit them to the main
repo in /public/img/

lefnire:
might be better in web repo than shared, but heck - "if aint broke..."

PR #45, opened 2013-11-28
suggested fix for adding "special" pets 

snicker:
so that the list does not have to be duplicated multiple times. the user model can be updated in `habitrpg` to pull from here instead of hard coding it

I couldn't actually test this, so I didn't want to merge straight away... could someone else verify this works? @paglias, @lefnire

### DIFF
--- a/dist/habitrpg-shared.js
+++ b/dist/habitrpg-shared.js
@@ -53,7 +53,7 @@ process.chdir = function (dir) {
 };
 
 },{}],2:[function(require,module,exports){
-var global=self;/**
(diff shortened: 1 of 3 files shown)


snicker:
Sorry about that one, it should be passing now!

paglias:
@snicker the code is ok, you just need to run `grunt` in order to compile the javascript file & commit it to github - great!

snicker:
Ahhhhhhhhhhhhhh

snicker:
Alrighty, this should all be peachy now. Merging!

lefnire:
pushed, thanks @snicker !

PR #54, opened 2013-12-14
Display friendly text instead of data element name when feeding pets

SabreCat:
Also incidentally fixes "Potatoe" typo.

Not locally tested... my personal Habit is broken again after Mounts rollout. But this syntax works elsewhere, so it should be OK?

### DIFF
--- a/tests/algos.mocha.coffee
+++ b/tests/algos.mocha.coffee
@@ -158,18 +158,28 @@ describe 'User', ->
       algos.score(user, user.dailys[0], 'up')
       expect(user.items.eggs).to.eql {}
       expect(user.items.hatchingPotions).to.eql {'Golden': 1}
+      expect(user.items.food).to.eql {}
 
     it 'gets a bear cub egg', ->
       sinon.stub(Math, 'random', cycle [0, 0.55])
       algos.score(user, user.dailys[0], 'up')
       expect(user.items.eggs).to.eql {'BearCub': 1}
       expect(user.items.hatchingPotions).to.eql {}
+      expect(user.items.food).to.eql {}
+
+    it 'gets honey', ->
+      sinon.stub(Math, 'random', cycle [0, 0.9])
+      algos.score(user, user.dailys[0], 'up')
+      expect(user.items.eggs).to.eql {}
+      expect(user.items.hatchingPotions).to.eql {}
+      expect(user.items.food).to.eql {'Honey': 1}
 
     it 'does not get a drop', ->
       sinon.stub(Math, 'random').returns 0.5
       algos.score(user, user.dailys[0], 'up')
       expect(user.items.eggs).to.eql {}
       expect(user.items.hatchingPotions).to.eql {}
+      expect(user.items.food).to.eql {}
 
     afterEach ->
       Math.random.restore()


SabreCat:
Haha, "the `master` branch can be safely deleted." Suuuure.

I realize now I just forgot to branch the way I usually do. Derp. Hopefully no harm done.

lefnire:
I did a force-push to master to undo the merge. Looks like it closed this ticket, something went haywire. I'd re-open it against develop. Incidentally, the commit won't work - `egg` at this point in the code is a String, not an Object. I'll revisit soon to fix

lefnire:
(glad you said something @SabreCat , I wouldn't have caught that)

SabreCat:
Ugh. Okay. This is no high priority, more a hotfix I was doing to get myself warmed up for working on algos and content, so I'll take another stab once I can get my local copy running again. (Every time we have any substantial update, it goes belly-up and I have to start from scratch, sigh.)

Maybe this time I can figure out how to get habitrpg-shared working in both npm and bower...

Apologies for the inconvenience, I should know better :\

lefnire:
What seems to be the problem? You wanna G+ and I can guide you through?

SabreCat:
Got some housework I'm doing (just checking in here during a Pomodoro break), so I'll have to take you up on that 
later. But for context, two recurring problems:
1. My local site breaks one way or another after most pulls. Sometimes doing npm and bower updates or installs is enough to fix it, but just as often, I get the "de-Angularized" page with all modals stacked atop one another, 500 errors, etc.
2. I seem to be able to get a local repo of habitrpg-shared running under habitrpg\node_modules, but not under habitrpg\node_modules\bower\node_modules. My next guess at it is going to be to muck with package.json in the bower directory.

SabreCat:
I think I might be best off trying for a remote instead of a local branch for actually running habitrpg-shared. Clone it locally outside the habitrpg folder structure, make changes, push to a sabe/something branch that I've pointed my local package.json files at.

lefnire:
1) Most big pushes (mounts, classes) are Schema changers. They're incompatible with the the old database snapshots, so you either have to (a) run the migration (if it's developed),  or (b) `db.dropDatabase()` and start anew. (b) is what I do up until I'm testing migrations
2) you mac or PC? if mac, you can do `npm link` and `ln -s` for npm/bower respectively. If windows, I'm not sure

I can show you on a G+ how I have things setup for that

SabreCat:
1. Ah, that makes a lot of sense. Hadn't thought of dropping the database.
2. PC, alas. Symlinking appears to straight-up not work on that platform, based on my own research into the problems I've had.

lefnire:
IIRC (and I could be wrong) `npm link` gets around symlinking on windows, I've heard some people  got it working. Evidently bower also has `bower link` ([info](https://github.com/bower/bower/issues/75#issuecomment-10627311)), which hopefully will have windows support as well. But it's probably a bit less likely, since npm has a huge windows developer base, where bower is only recently becoming defacto

SabreCat:
Wait, do we even have a habitrpg-shared under bower/node_modules anymore? I could swear that used to be a thing (the wiki page on working with habitrpg-shared talks about it), but now I can't even find it.

lefnire:
it should be under both. The server uses the coffee files in node_modules/habitrpg-shared/scripts, and the client use the public/bower_components/habitrpg-shared/dist/habitrpg-shared.js

SabreCat:
OK, I don't even know what's going on here then, haha. Back to my prior plan of wiping out my clones and starting over--hopefully I can rig up something sustainable this time. Thanks for helping me wrap my mind around it!

PR #58, opened 2013-12-21
Content update to match habitrpg/#2019 fix

SabreCat:
Removing remaining religious references from Healer equipment. "Mantle" instead of "Vestment", "Scepter" instead of "Crosier", "Physician" instead of "Priest", "Protector" instead of "Hospitaler".

### DIFF
--- a/script/content.coffee
+++ b/script/content.coffee
@@ -437,10 +437,10 @@ _.each api.hatchingPotions, (pot,k) ->
 api.food =
   Meat:             text: 'Meat', target: 'Base'
   Milk:             text: 'Milk', target: 'White'
-  Potatoe:          text: 'Potato', target: 'Desert'
-  Strawberry:       text: 'Strawberry', target: 'Red'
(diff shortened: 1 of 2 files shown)


SabreCat:
e5a9825: Tightens up wordings on food drop notifications. Where before we had the rather awkward "You've found a Fish Food!" (which sounds like you got a canister of betta flakes), it'll now say simply "You've found a Fish!" Where before we referred to uncountable objects with an article, "You've found a Honey" (like, a romantic partner?), it now simply says "You've found Honey!"

I still want to figure out how to make similar improvements to the feeding text, but it turns out it's not as easy as I previously thought.

SabreCat:
Hmm, e5a9825 produces some oddments in the Market. Working on a better implementation, but need to do some testing.

SabreCat:
Still testing, hold off merge for now.

SabreCat:
Ugh. Looks like I just broke things. Going to split this out into a separate pull request for the healer thing (which should be fine) and rebranch to mess with foods.

PR #61, opened 2013-12-22
Fix preening bug: null values

djuretic:
There was a bug in the preening user history code: it was storing null values. Sadly, that means that there are loss of historical data.
@lefnire, with this change merged, all the tests are going to pass :)


lefnire:
omg! you did it?? you fixed all the tests, I have been dreading them (since classes was such a huge overhaul). thanks man!

PR #62, opened 2013-12-22
Improve food drop notification phrasing

SabreCat:
Makes notifications of food drops sound more like natural English.

"You found a Milk Food!" becomes "You found Milk!"
"You found a Fish Food!" becomes "You found a Fish!"

...and so forth.


SabreCat:
It turns out the previous commits I attempted were fine (though it did have a bug that I fixed in fce23fc), I just didn't have any way of testing them while my local repo was all busted. Seems to check out smooth now, though.

PR #66, opened 2013-12-26
Fix HabitRPG/habitrpg#2100 (can't get Ultimate Gear with Mage stuff)

SabreCat:
Mage (currently Wizard) can now get Ultimate Gear achievement for picking up all the Royal Magus equipment. Code only requires max shield if the user does not have a two-handed weapon equipped.

Tested that Wizard gets the achievement upon purchasing final item, and that other classes (used Warrior) still need to get shields to earn their badge.


lefnire:
perfect

PR #67, opened 2013-12-31
Add 5 new hair and beard colors

phillipthelen:
While habitrpg hase a rather large range of different skin colors, the hair color choices are rather limited. Here are 5 new hair and beard colors (bright red, green, purple, pink and turquoise).

I hope I did everything correct and the changes can be merged like this. if not, I would be thankfull for help on what I have to do different


Pandoro:
There is a BIG new set of hair colors in the [pixel thread](https://github.com/HabitRPG/habitrpg/issues/946) which is still in the pipeline, are these the same or did you make them in parallel?

phillipthelen:
As i didn't know of the pixelart thread, i did these in parallel. My mistake. As the ones there are better/more i'll close this. Sorry for the disturbance.

Pandoro:
No problem! Sorry that this happened :( Still great you did it :D Shows the
love people have for habit :)

PR #70, opened 2014-01-02
Add Critical Hit notificaiton via user._tmp

### LINKED ISSUE #2029: Critical hits need their own alert callout
For a little while after the Classes rollout, I was baffled at why sometimes I was getting upwards of 6 gold for a single click, and other times I got only 2 or so. My hunch at this point (after the recent PER rebalancing) is that critical hits are happening sometimes? Do we display an alert "Critical hit!" or similar to let the user know what just happened?

Some cool animations would be awesome, but at least we need an alert, heh.

colegleason:
Adds the critical ht bonus amount to user._tmp.crit so that the client can notify the user if they get a crit.

Fixes HabitRPG/habitrpg#2029.

### DIFF
--- a/script/index.coffee
+++ b/script/index.coffee
@@ -761,6 +761,8 @@ api.wrap = (user) ->
       addPoints = ->
         # ===== CRITICAL HITS =====
         _crit = user.fns.crit()
+        # if there was a crit, alert the user via notification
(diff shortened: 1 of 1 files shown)


lefnire:
There we go! Simpler than I thought it'd be. Now we just need an angular $watch

PR #71, opened 2014-01-03
Tasks get blue when you do them often, and the text for Pickpocket should reflect this

JasonGross:
I believe the sequence is red -> yellow -> green -> blue.  The text for 'Pickpocket' now reflects this slightly better, I believe.


lefnire:
Fair 'nuff!

SabreCat:
This applies to Burst of Flames as well, methinks.

PR #73, opened 2014-01-05
fix(?) for issue 2222 - untick removes too little XP,GP

Alys:
This is a proposed fix for issue 2222 ("unticking a todo with a checklist removes much less XP and GP than it should") BUT I HAVE NOT TESTED THIS (I don't have a local install yet - sorry). Please see my comments at the end of that issue. Note that I don't know why "if direction is 'up'" was originally included - there might have been (and still be) a good reason for it.

### DIFF
--- a/script/index.coffee
+++ b/script/index.coffee
@@ -743,7 +743,7 @@ api.wrap = (user) ->
             if direction is 'down' and task.type is 'daily' and options.cron
               nextDelta *= (1 - _.reduce(task.checklist,((m,i)->m+(if i.completed then 1 else 0)),0) / task.checklist.length)
             # If To-Do, point-match the TD per checklist item
-            if direction is 'up' and task.type is 'todo'
+            if task.type is 'todo'
               nextDelta *= task.checklist.length
 
           unless task.type is 'reward'


lefnire:
What this will do is, a user will gain points from checking a todo no matter if they're checking or unchecking. Alas

Alys:
Forgive me for arguing - especially if I'm still wrong! - but I don't think this will cause the user to gain points if they are unchecking. The index.coffee script already contains this code (line 738):
          nextDelta = Math.pow(0.9747, currVal) \* (if direction is 'down' then -1 else 1)
which is already being executed for any todo that's either ticked or unticked. Note the direction +/- 'if' clause which ensure that points are subtracted if the user is unchecking.

The trouble at the moment is that if the todo contains a checklist, there is further code that multiplies nextDelta by the checklist length:
          nextDelta *= task.checklist.length
This is currently only being executed when you complete a todo (direction is 'up'). If we also execute it when you untick a todo (direction is 'down') then nextDelta will still be negative and so points will still be subtracted.

"the math used to determine the amount of points going up vs points going down won't give you the same results after modifying the task.value" - Yes, but the difference seems to be small when there is no checklist (e.g., issue #2221 - 1.2 GP versus 1.1 GP). When there is a checklist, the multiplying factor of every checklist item makes the difference much larger (7.3 GP versus 3.4 GP for only two checklist items). Including the task.checklist.length multipler for unticking todos will reduce the difference to a much smaller amount.

lefnire:
Ah! Interesting, let me give it a whirl and sorry for the quick dismiss

Alys:
I've just noticed there's a completely separate issue with MP being ADDED instead of subtracted when you untick a todo. That's handled by different code (I've already found where). I'll be submitting a separate fix for that, in a few minutes hopefully.

EDIT: Fix submitted. https://github.com/HabitRPG/habitrpg-shared/pull/76

PR #75, opened 2014-01-05
Rebirth fixes: deequip pets and mounts, allow stacking at max level

SabreCat:
Two tweaks to Rebirth behavior:
- Undergoing a Rebirth deequips the user's current pet and mount, if any.
- In the unlikely, hardest-core event that a player reaches level 100 more than once, they can stack Rebirths there as many times as they make it. (Previously they needed to exceed their prior level to re-earn the achievement, making it impossible to stack the Rebirth achievement after being reborn from lvl 100.)

### DIFF
--- a/script/index.coffee
+++ b/script/index.coffee
@@ -435,6 +435,8 @@ api.wrap = (user) ->
         gear[type].weapon = 'weapon_warrior_0'
         gear[type].head   = 'head_base_0'
         gear[type].shield = 'shield_base_0'
+      if user.items.currentPet then user.ops.equip({params:{type: 'pet', key: user.items.currentPet}})
+      if user.items.currentMount then user.ops.equip({params:{type: 'mount', key: user.items.currentMount}})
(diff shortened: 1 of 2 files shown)


script/index.coffee
```diff
@@ -449,7 +451,7 @@ api.wrap = (user) ->
       if not (user.achievements.rebirths)
         user.achievements.rebirths = 1
         user.achievements.rebirthLevel = lvl
-      else if (lvl > user.achievements.rebirthLevel)
+      else if (lvl > user.achievements.rebirthLevel or lvl = 100)
```
lefnire: this will assign `lvl` to 100. Might be intended lil' trick (I have a weakness for tricks!), but just in case you didn't intend it: `lvl == 100` or `lvl is 100`

SabreCat: Whoops, darn that assignment operator! My chump QBasic coding background keeps betraying me. No, I did indeed mean `is`, you can fix that before merging.

PR #76, opened 2014-01-05
subtract MP when unticking a todo (instead of add)

Alys:
Currently, when you untick a completed todo, you GAIN MP. Lines 837-839 of this change will cause you to loose it instead.

Line 841 will prevent your MP becoming negative if you used up all your MP before unticking the todo... BUT do we want to prevent negative MP under that circumstance? Perhaps it should be allowed to go negative? The circumstance I'm thinking of is:
1. You tick off a todo, causing your MP to reach (for example) 10.
2. You cast a spell that costs 10 MP. MP is now 0.
3. You realise that you had ticked off the todo accidentally (or deliberately to rort the system!) and so you untick the todo, which (without line 841) would cause your MP to go to -1. Strictly speaking, it SHOULD be -1 because you had spent MP that you hadn't truly earned, but negative MP might look weird and cause "bug" reports. Line 841 will prevent it being negative.

### DIFF
--- a/script/index.coffee
+++ b/script/index.coffee
@@ -834,8 +834,11 @@ api.wrap = (user) ->
           else
             calculateDelta()
             addPoints() # obviously for delta>0, but also a trick to undo accidental checkboxes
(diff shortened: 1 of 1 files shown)


lefnire:
nice, you're on fire! I think you're right, the current solution is the lesser of two evils. It allows them to cheat (even if accidentally), but we're gonna see much fewer bug reports as a result. Habit being a not-so-regulated in the cheating department app as is & all

PR #79, opened 2014-01-07
Server side of Rebirth-related fix: don't remove Wolf Eggs upon drops unlock

SabreCat:
Previously, if user had >1 Wolf Egg, performed a Rebirth, then unlocked drops again, they would be reset to 1 Wolf Egg. Now Habit checks to see if they have some Wolf Eggs already, and if so, gives them one instead of setting the quantity to one.

Question for @lefnire / @paglias: I'm being careful about not performing a `++` on an undefined quantity. Is that the right code practice, or will JS turn an undefined into a 1 if the type is number and I add to it?

### DIFF
--- a/dist/habitrpg-shared.js
+++ b/dist/habitrpg-shared.js
@@ -63,7 +63,7 @@ process.chdir = function (dir) {
 };
 
 },{}],3:[function(require,module,exports){
-var global=self;/**
+var global=typeof self !== "undefined" ? self : typeof window !== "undefined" ? window : {};/**
  * @license
  * Lo-Dash 2.4.1 (Custom Build) <http://lodash.com/>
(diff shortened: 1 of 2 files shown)


paglias:
yes, it's correct to check if it's a number.

```
var a = undefined;
a++
a === NaN (not a number)
```

lefnire:
I hate this stuff, man

paglias:
oh my god, it's horrible

SabreCat:
lolwut? So if I added null to it first, it'd work? xD

lefnire:
@SabreCat [funny you should say "wut"](https://www.destroyallsoftware.com/talks/wat)

SabreCat:
@lefnire Oh Primes. I laughed so hard. My fiancée probably thinks I'm such a dork xD

PR #83, opened 2014-01-08
Fixed Ski-sassin Pole

Goldenwyrm:
Rogue's shield should increases STR attribute.


Goldenwyrm:
Probably all rogues that bought it should have a fix of their attributes.

lefnire:
it actually should be `str: 8, value: 90` (updated the code) rogue off-hands are kinda funky in that they have one more tier than other class's shields, and these winter items should match each class's 2nd-best equiv. Thanks for the PR though, sorry bout that!

PR #84, opened 2014-01-08
Custom Day Start: change condition on when we run cron

colegleason:
This is an attempt to fix some of the problems people have with custom
day start, especially those in habitrpg/habitrpg#1057.

I believe this is a bug in when we decide to run cron. Right now we look
for the difference between `startOfDay(user.lastCron)` and `now` to be
greater than 0. I'll explain why this is a problem.

For the following cases, let today be January 8th and yesterday be January 7th.

Consider the scenario where `dayStart = 0` and `user.lastCron` is sometime yesterday.

`lastCron.start('days').add('d', dayStart)` will just be January 7th at midnight. Then
the difference will be greater than 0 anytime after midnight on January
8th. Works as expected.

However, consider the case where `dayStart` is set to something greater than
0, like 3. Now let's say the user logs on after midnight on January 8th
but before their Custom Day Start (something common in many reports),
and cron is run.

`lasCron.start('days').add('d', dayStart)` will be January 8th at
3am. Therefore the difference won't be greater than 0 until January 9th
at 3am and their dailies, etc won't become unchecked.

Therefore, this commit adds two extra conditions to check when `daysMissed`
is equal to zero:
- cronBeforeStart: Was the last cron run before the Custom Day Start
  time for today's date?
- nowAfterStart: Have we passed the Current Day Start time for today's
  date?

I believe that both of these condition must be true OR daysMissed must
be greater than zero to run cron. Please, please, please double and
triple check my logic here. I don't want to make this any more of a
mess, because working with times makes my brain hurt.

(cc @lefnire)

### DIFF
--- a/script/index.coffee
+++ b/script/index.coffee
@@ -31,7 +31,10 @@ api.startOfWeek = api.startOfWeek = (options={}) ->
 
 api.startOfDay = (options={}) ->
   o = sanitizeOptions(options)
-  moment(o.now).startOf('day').add('h', o.dayStart)
+  dayStart = moment(o.now).startOf('day').add('h', o.dayStart)
+  # if between midnight and Custom Day Start, jump back to previous day
+  dayStart.subtract('day', 1) if moment(o.now).isBefore(dayStart)
+  return dayStart
 
 dayMapping = {0:'su',1:'m',2:'t',3:'w',4:'th',5:'f',6:'s'}
 
@@ -1239,4 +1242,3 @@ api.wrap = (user, main=true) ->
     get: ->
       tasks = user.habits.concat(user.dailys).concat(user.todos).concat(user.rewards)
       _.object(_.pluck(tasks, "id"), tasks)
-
--- a/tests/algos.mocha.coffee
+++ b/tests/algos.mocha.coffee
@@ -487,6 +487,8 @@ describe 'Helper', ->
     expect(shared.startOfDay({now: new Date(2013, 0, 1, 0)}).format('YYYY-MM-DD HH:mm')).to.eql '2013-01-01 00:00'
     expect(shared.startOfDay({now: new Date(2013, 0, 1, 5)}).format('YYYY-MM-DD HH:mm')).to.eql '2013-01-01 00:00'
     expect(shared.startOfDay({now: new Date(2013, 0, 1, 23, 59, 59)}).format('YYYY-MM-DD HH:mm')).to.eql '2013-01-01 00:00'
+    # between midnight and custom day start
+    expect(shared.startOfDay({now: new Date(2013, 0, 2, 2), dayStart: 4}).format('YYYY-MM-DD HH:mm')).to.eql '2013-01-01 04:00'
 
   it 'counts pets', ->
     pets = {}
--- a/tests/algos.mocha.coffee
+++ b/tests/algos.mocha.coffee
@@ -46,7 +46,7 @@ rewrapUser = (user)->
 expectStrings = (obj, paths) ->
   _.each paths, (path) -> expect(obj[path]).to.be.ok()
 
-# options.daysAgo: days ago when the last cron was executed
+# options.hoursAgo: hours since the last cron was executed
 beforeAfter = (options={}) ->
   user = newUser()
   [before, after] = [user, _.cloneDeep(user)]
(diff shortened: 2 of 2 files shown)


colegleason:
@lefnire, what is `api.startOfDay` supposed to return? I believe our underlying issue for most reports in habitrpg/habitrpg#1057 may be that it is either incorrect or that it is being used incorrectly in the code.

Example (as above):
Let `dayStart = 3`.
1. User logs on before midnight local time on January 7th. `api.startOfDay now` returns a date object representing January 7th at 3:00am.
2. User logs on between midnight and 3:00am on January 8th. `api.startOfDay now` returns a date object representing January 8th at 3:00am.
3. User logs on after 3:00am on January 8th. `api.startOfDay now` returns a date object representing January 8th at 3:00am.

I believe in (2) the intent was for it to return a date object representing January 7th at 3:00am, thus making all calculations using 'moments().diff`or`shouldDo` work from the previous day's start time. 

If that is the case then I'll trash this commit and fix `api.startOfDay`. I just want to make sure that is the intent of the function and how it is being used elsewhere.

lefnire:
That's exactly the intent. The thing is, we wrote quite a few [tests here](https://github.com/HabitRPG/habitrpg-shared/blob/develop/tests/algos.mocha.coffee#L382) in order expect that #2 issue possibility, but maybe it's not kicking. But yes, startofDay is meant to return the proper window of time given the user's timezone and preferences.

Take a look at those tests to see if they give you any leads. We could do a pair-session sometime if you're down, no pressure if not!

colegleason:
I'll take a look at those tests. If you pull [this commit](https://github.com/colegleason/habitrpg-shared/commit/9d75c3bfd6deb8142a7ca9ea77eb5c53a06c54bc) and run npm test, though, you'll see we have some sort of problem here.

colegleason:
I just rebased, so you can pull that entire branch and it should just
include the tests.

On Wed, Jan 8, 2014 at 5:19 PM, Tyler Renelle notifications@github.comwrote:

> That's exactly the intent. The thing is, we wrote quite a few tests herehttps://github.com/HabitRPG/habitrpg-shared/blob/develop/tests/algos.mocha.coffee#L382in order expect that
> #2 https://github.com/HabitRPG/habitrpg-shared/pull/2 issue
> possibility, but maybe it's not kicking. But yes, startofDay is meant to
> return the proper window of time given the user's timezone and preferences.
> 
> Take a look at those tests to see if they give you any leads. We could do
> a pair-session sometime if you're down, no pressure if not!
> 
> —
> Reply to this email directly or view it on GitHubhttps://github.com/HabitRPG/habitrpg-shared/pull/84#issuecomment-31887147
> .

## 

## Cole Gleason

Email: cg@colegleason.com
Website: colegleason.com

lefnire:
Heading out, will pull and investigate tonight / tm morn

colegleason:
@lefnire I think this is the fix, perhaps? Still digging through the tests because the whole runCron, cronMatrix stuff is confusing.

script/index.coffee
```diff
@@ -31,7 +31,10 @@ api.startOfWeek = api.startOfWeek = (options={}) ->
 
 api.startOfDay = (options={}) ->
   o = sanitizeOptions(options)
-  moment(o.now).startOf('day').add('h', o.dayStart)
+  dayStart = moment(o.now).startOf('day').add('h', o.dayStart)
+  # if between midnight and Custom Day Start, jump back to previous day
+  dayStart.subtract('day', 1) if moment(o.now).isBefore(dayStart)
```
lefnire, line 36: Just trying to wrap my head around things, did some quick tests on live:
1. Set CDS to 12
2. Knock me back a day (`db.users.update({_id:ID},{$set:{lastCron:+new Date('01/08/2014')}})`)
3. Sync
4. Cron is run
   - Cron shouldn't have run, hadn't yet hit CDS (this was at 11:40).

In cron, `daysMissed = api.daysSince user.lastCron, _.defaults({now}, user.preferences)`

Now up here we say `dayStart = moment(o.now).startOf('day').add('h', o.dayStart)` (start of lastCron's day, aka yesterday at 12)

`dayStart.subtract('day', 1) if moment(o.now).isBefore(dayStart)` says "if _that's_ before yesterday at 12, minus another day". So now I'm 2 days behind, even though it's set to yesterday.

Or... maybe that's accurate, given CDS? If I logged in before yesterday before 12, that means I'm technically 2 days behind?

lefnire, line 36: No wait, you're right. I would technically be 2 days behind then.

PR #85, opened 2014-01-09
Rebirth fix: flip owned items to explicit false 

SabreCat:
Per current handling of `canOwn`, gear items set to an explicit `false` can be rebought after loss. This commit makes the Rebirth op handle owned gear this way, so a Reborn character doesn't unfairly lose access to limited edition equipment.

### DIFF
--- a/script/index.coffee
+++ b/script/index.coffee
@@ -441,8 +441,9 @@ api.wrap = (user, main=true) ->
           gear[type].shield = 'shield_base_0'
         if user.items.currentPet then user.ops.equip({params:{type: 'pet', key: user.items.currentPet}})
         if user.items.currentMount then user.ops.equip({params:{type: 'mount', key: user.items.currentMount}})
-        # Strip owned gear down to the training sword
-        gear.owned = {weapon_warrior_0:true}
+        # Strip owned gear down to the training sword, but preserve purchase history so user can re-purchase limited edition equipment
+        _.each gear.owned, (v, k) -> if gear.owned[k] then gear.owned[k] = false; true
+        gear.owned.weapon_warrior_0 = true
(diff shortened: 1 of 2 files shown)


script/index.coffee
```diff
@@ -441,8 +441,9 @@ api.wrap = (user, main=true) ->
           gear[type].shield = 'shield_base_0'
         if user.items.currentPet then user.ops.equip({params:{type: 'pet', key: user.items.currentPet}})
         if user.items.currentMount then user.ops.equip({params:{type: 'mount', key: user.items.currentMount}})
-        # Strip owned gear down to the training sword
-        gear.owned = {weapon_warrior_0:true}
+        # Strip owned gear down to the training sword, but preserve purchase history so user can re-purchase limited edition equipment
+        _.each gear.owned, (v, k) -> if gear.owned[k] then gear.owned[k] = false; true
```
lefnire, line 445: Can also do `gear.owned = _.mapValues gear.owned, ->false` (why do you put up with me?)

SabreCat, line 445: Naw, all good! I don't always grok what all the Lo-Dash bits are useful for, even with their examples. Your suggestion does one better because it spares that unused `v`, for starters.

lefnire, line 445: Woops, just realized you're doing it this way to account for `undefined`s. There shouldn't be any `undefined`s in `gear.owned` (if ever something's _really_ removed, it's done so with `$unset` or `delete` - but gear isn't really removed as such in the app); however, your approach is "better safe than sorry", so :+1:

SabreCat, line 445: I happened into good practice somewhat by accident, then. I know at least one of my attempts set _every possible gear item_ to `false` xD

PR #89, opened 2014-01-10
Calculate ultimateGear achievement in user.ops

colegleason:
This does several things:
- Moves the calculation of the ultimateGear achievement into its on
  function on user.ops.
- Calls user.ops.ultimateGear on user.save instead of on item purchase
- Fixes bug where ultimateGear was not awarded if user wasn't wearing
  their class gear at the time of item purchase.
- Adds content.classes and content.gearTypes which we should start using
  elsewhere.

Fixes habitrpg/habitrpg#2232.

### DIFF
--- a/dist/habitrpg-shared.js
+++ b/dist/habitrpg-shared.js
@@ -11884,11 +11884,7 @@ var process=require("__browserify_process");(function() {
               message = "Bought " + item.text + ".";
             }
             if (!user.achievements.ultimateGear && item.last) {
-              user.ops.ultimateGear({
-                params: {},
-                query: {},
-                body: {}
-              });
+              user.fns.ultimateGear();
             }
           }
           user.stats.gp -= item.value;
@@ -12241,38 +12237,6 @@ var process=require("__browserify_process");(function() {
             cb(null, user);
           }
           return delta;
-        },
-        ultimateGear: function(req, cb) {
-          var gear, lastGearClassTypeMatrix, ownedLastGear, shouldGrant;
-
-          gear = typeof window !== "undefined" && window !== null ? user.items.gear.owned : user.items.gear.owned.toObject();
-          ownedLastGear = _.chain(content.gear.flat).pick(_.keys(gear)).values().filter(function(gear) {
-            return gear.last;
-          });
-          lastGearClassTypeMatrix = {};
-          _.each(content.classes, function(klass) {
-            lastGearClassTypeMatrix[klass] = {};
-            return _.each(content.gearTypes, function(type) {
-              lastGearClassTypeMatrix[klass][type] = false;
-              return true;
-            });
-          });
-          ownedLastGear.each(function(gear) {
-            if (gear.twoHanded) {
-              lastGearClassTypeMatrix[gear.klass]["shield"] = true;
-            }
-            return lastGearClassTypeMatrix[gear.klass][gear.type] = true;
-          });
-          shouldGrant = _(lastGearClassTypeMatrix).values().reduce((function(ans, klass) {
-            return ans || _(klass).values().reduce((function(ans, gearType) {
-              return ans && gearType;
-            }), true);
-          }), false).valueOf();
-          user.achievements.ultimateGear = shouldGrant;
-          if (typeof cb === "function") {
-            cb(null, user);
-          }
-          return shouldGrant;
         }
       };
     }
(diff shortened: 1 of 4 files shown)


colegleason:
Totally a bug in this. Let me fix that real quick.

colegleason:
There we go, that sped things up. What do you think?

I also want to start breaking things out of index.coffee if possible. The file is rather unruly.

lefnire:
We actually were originally in multiple files (view helpers, algorithms, etc) and had too many problems with circular dependencies + browserify over the course of a few months. Tried many solutions, just wasn't worth it in the end and had to bring them all into one file. `content.coffee` is the only one I kept intact, but as you can see it redefines some functions (`$w()`, `crit`, etc) due to circ dep issues.

Anyway, i'll check out this PR soon as I can. I might write up a quick algos.mocha.coffee test for it monday just in case, then merge it Monday. Thanks for all the hard work man! Sorry I'm not able to merge / review much this weekend

colegleason:
No problem. And I should really be writing the tests, I just haven't gotten
to it yet.

On Fri, Jan 10, 2014 at 2:50 PM, Tyler Renelle notifications@github.comwrote:

> We actually were originally in multiple files (view helpers, algorithms,
> etc) and had too many problems with circular dependencies + browserify over
> the course of multiple months. Tried many solutions, just wasn't worth it
> in the end and had to bring them all into one file. content.coffee is the
> only one I kept intact, but as you can see it redefines some functions (
> $w(), crit, etc) due to circ dep issues.
> 
> Anyway, i'll check out this PR soon as I can. I might write up a quick
> algos.mocha.coffee test for it monday just in case, then merge it Monday.
> Thanks for all the hard work man! Sorry I'm not able to merge / review much
> this weekend
> 
> —
> Reply to this email directly or view it on GitHubhttps://github.com/HabitRPG/habitrpg-shared/pull/89#issuecomment-32064716
> .

## 

## Cole Gleason

Email: cg@colegleason.com
Website: colegleason.com

colegleason:
Rebased to make this merge-able and squashed all commits into one.

lefnire:
Some concerns about this commit, bear with me:

> Moves the calculation of the ultimateGear achievement into its on function on user.ops.

user.ops are handled specially. If an op is called client-side, it's then replicated server-side. From "Calls user.ops.ultimateGear on user.save instead of on item purchase", this might mean it will get called twice. Sure we don't want it in user.fns?

> Calls user.ops.ultimateGear on user.save instead of on item purchase

A bit concerned that the original 1-liner, now a 30-liner, is called for every possible operation (score, shirt-change, etc). In many cases, "meh" - but in this case, it looks like there's quite a lot getting computed. Sure we don't want this only on purchase?

> Fixes bug where ultimateGear was not awarded if user wasn't wearing their class gear at the time of item purchase.

:+1: 

> Adds content.classes and content.gearTypes which we should start using elsewhere.

:+1:

colegleason:
Sorry, it actually isn't in user save anymore (or shouldn't be). I forgot
to modify the commit message. It should only process on buy.

Not sure about user ops versus fns. I thought we would want it on client
and server

lefnire:
So, `user.fns` and `user.ops` are both available client & server-side. `user.fns` is helper functions on the user object, avail both C&S. But here's what happens with `user.ops`:

Anything inside ops, if it's called on the client it will run the function, then it will send a `POST /user/batch-update {op:OP_NAME}`. The server will receive that op, run the function, and save the user. So for example, `user.ops.score` will score the task in the browser, `POST /api/v2/user/batch-update [{op:"score",params:...}]`, server will run `user.ops.score()` and then `user.save()`.

So if achievements is handled in ops, and in user.save / user.buy, it will get duplicated.

colegleason:
Ok, so if buy is an op, then it is called on client and server when we buy
an item. We only want ultimateGear to run once on client and once on
server, so it should be a fn instead, correct?

IIRC it was only being called once on the server, but that may be because
of the check to see if you already have the achievement.

So, user.fns and user.ops are both available client & server-side.
user.fnsis helper functions on the user object, avail both C&S. But
here's what
happens with user.ops:

Anything inside ops, if it's called on the client it will run the function,
then it will send a POST /user/batch-update {op:OP_NAME}. The server will
receive that op, run the function, and save the user. So for example,
user.ops.score will score the task in the browser, POST
/api/v2/user/batch-update [{op:"score",params:...}], server will run
user.ops.score() and then user.save().

So if achievements is handled in ops, and in user.save / user.buy, it will
get duplicated.

—
Reply to this email directly or view it on
GitHubhttps://github.com/HabitRPG/habitrpg-shared/pull/89#issuecomment-32691968
.

colegleason:
@lefnire, Made the changes and tested.

PR #91, opened 2014-01-11
Fix broken test

colegleason:
`yesterday` was actually two days ago since it had already been run through `startOfDay`.


colegleason:
I don't think this is valid anymore.

PR #94, opened 2014-01-18
fix(perfect-day): do not count non-active dailies when calculating perfect-day

### LINKED ISSUE #2446: Perfect day achievement : not awarded if you have grey dailies uncomplete
Hi there,

I removed a daily yesterday (without checking it before), and had my whole list completed at the end of the day. I didn't received any achievement for this "perfect day".

Is this intended ? (it would make sense, in a way)

If so, would checking the daily before removing it prevent this behaviour ?

Or maybe it's not related at all, and linked to something else ?

Regards

Aiseant

colegleason:
When we calculate the perfect-day achievement, we should check to see if
the daily had repeat set for yesterday as well as if it was completed.

Fixes HabitRPG/habitrpg#2446

### DIFF
--- a/script/index.coffee
+++ b/script/index.coffee
@@ -33,7 +33,7 @@ api.startOfDay = (options={}) ->
   o = sanitizeOptions(options)
   moment(o.now).startOf('day').add('h', o.dayStart)
 
-dayMapping = {0:'su',1:'m',2:'t',3:'w',4:'th',5:'f',6:'s'}
+api.dayMapping = {0:'su',1:'m',2:'t',3:'w',4:'th',5:'f',6:'s'}
 
 ###
   Absolute diff from "yesterday" till now
@@ -48,13 +48,13 @@ api.daysSince = (yesterday, options = {}) ->
 api.shouldDo = (day, repeat, options={}) ->
   return false unless repeat
   o = sanitizeOptions options
-  selected = repeat[dayMapping[api.startOfDay(_.defaults {now:day}, o).day()]]
+  selected = repeat[api.dayMapping[api.startOfDay(_.defaults {now:day}, o).day()]]
   return selected unless moment(day).zone(o.timezoneOffset).isSame(o.now,'d')
   if options.dayStart <= o.now.hour() # we're past the dayStart mark, is it due today?
     return selected
   else # we're not past dayStart mark, check if it was due "yesterday"
     yesterday = moment(o.now).subtract(1,'d').day() # have to wrap o.now so as not to modify original
-    return repeat[dayMapping[yesterday]] # FIXME is this correct?? Do I need to do any timezone calcaulation here?
+    return repeat[api.dayMapping[yesterday]] # FIXME is this correct?? Do I need to do any timezone calcaulation here?
 
 
 ###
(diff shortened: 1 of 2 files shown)


colegleason:
Quick question here: Should we only calculate perfect days id daysMissed == 1?

tests/algos.mocha.coffee
```diff
@@ -159,6 +159,14 @@ describe 'User', ->
     expect(user.stats.buffs.str).to.be 1
     expect(user.achievements.perfect).to.be 1
 
+    # Handle greyed-out dailys
+    yesterday = moment().subtract('days',1);
+    user.dailys[0].repeat[shared.dayMapping[yesterday.day()]] = 0
+    _.each user.dailys[1..], (d)->d.completed = true
```
lefnire, line 165: What'ya mean you don't like coffee, you're whipping up all sorts of tricks :)

colegleason, line 165: Don't get me wrong, I like the sugar. I guess it just takes a while to get used to. I'm also always fighting with my editor while writing it.

colegleason, line 165: I've done a lot of python, and It's pretty similar. Overall I would just prefer explicit returns and braces over indentation.

lefnire:
> Quick question here: Should we only calculate perfect days id daysMissed == 1?

I wonder if a simpler solution to this problem, and maybe the answer to the above question, is:

``` diff
diff --git a/script/index.coffee b/script/index.coffee
index a5d97bf..6ca5752 100644
--- a/script/index.coffee
+++ b/script/index.coffee
@@ -1144,14 +1144,8 @@ api.wrap = (user, main=true) ->
       user.stats.mp = user._statsComputed.maxMP if user.stats.mp > user._statsComputed.maxMP

       # "Perfect Day" achievement for perfect-days
-      lvlDiv2 = Math.ceil(user.stats.lvl/2)
-      clearBuffs =
-        if _.find(user.dailys, (d)->!d.completed) or user.preferences.sleep is true
-          {str:0,int:0,per:0,con:0,stealth:0,streaks:false}
-        else
-          user.achievements.perfect ?= 0
-          user.achievements.perfect++
-          {str:lvlDiv2,int:lvlDiv2,per:lvlDiv2,con:lvlDiv2,stealth:0,streaks:false}
+      perfect = true
+      clearBuffs = {str:0,int:0,per:0,con:0,stealth:0,streaks:false}

       # User is resting at the inn. Used to be we un-checked each daily without performing calculation (see commits before fb29e35)
       # but to prevent abusing the inn (http://goo.gl/GDb9x) we now do *not* calculate dailies, and simply set lastCron to today
@@ -1178,6 +1172,7 @@ api.wrap = (user, main=true) ->
               thatDay = moment(now).subtract('days', n + 1)
               scheduleMisses++ if api.shouldDo(thatDay, repeat, user.preferences)
           if scheduleMisses > 0
+            perfect = false
             delta = user.ops.score({params:{id:task.id, direction:'down'}, query:{times:scheduleMisses, cron:true}});
             user.party.quest.progress.down += delta if type is 'daily'

@@ -1211,7 +1206,13 @@ api.wrap = (user, main=true) ->
       user.fns.preenUserHistory()
       user.markModified? 'history'
       user.markModified? 'dailys' # covers dailys.*.history
-      user.stats.buffs = clearBuffs
+      user.stats.buffs =
+        if perfect
+          user.achievements.perfect ?= 0
+          user.achievements.perfect++
+          lvlDiv2 = user.stats.lvl/2
+          {str:lvlDiv2,int:lvlDiv2,per:lvlDiv2,con:lvlDiv2,stealth:0,streaks:false}
+        else clearBuffs

       # After all is said and done, progress up user's effect on quest, return those values & reset the user's
       progress = user.party.quest.progress; _progress = _.cloneDeep progress

```

lefnire:
Well, maybe not the answer actually... I think both our solutions effectively calculate perfect++ as "having nailed yesterday"

colegleason:
Almost, I think. That would fail if you missed a todo though.

colegleason:
So if you nailed a day 2 days ago but didn't log on yesterday, you don't get the buffs but you should get the achievement.

Do we calculate this on cron for perf reasons? Otherwise we could calculate it on every daily up score.

lefnire:
Ah, your'e correct. [Updated diff](https://gist.github.com/lefnire/8486480). (posting as diff in case we merge yours instead)

Well... once we merge your cron-script stuff (which I plan to), that shouldn't be an issue for either of our scenarios

colegleason:
Is that better?

script/index.coffee
```diff
@@ -1203,7 +1198,13 @@ api.wrap = (user, main=true) ->
       user.fns.preenUserHistory()
       user.markModified? 'history'
       user.markModified? 'dailys' # covers dailys.*.history
-      user.stats.buffs = clearBuffs
+      user.stats.buffs =
+        if perfect
+          user.achievements.perfect ?= 0
+          user.achievements.perfect++
+          lvlDiv2 = Math.ceil(user.stats.lvl/2)
```
colegleason, line 1205: @lefnire You only had user.stats.lvl/2 here but the tests indicated you wanted that rounded up.

lefnire, line 1205: Good catch!

lefnire:
Perfect man, thanks a million!

PR #95, opened 2014-01-18
Undo Feature

lefnire:
**WIP**. This implements undo by preventing the last action in the sync queue from pushing until either (a) new op pushed to queue, (b) 10s timeout, (c) window.onbeforeunload. See https://github.com/HabitRPG/habitrpg/tree/undo

I've played around with it and I think it's a simple & elegant approach, but it's definitely hitting some bugs. Likely need to shuffle around some stuff in userServices.js. Not ready for merge, but @colegleason wanted to bounce this approach off you

### DIFF
--- a/script/userServices.js
+++ b/script/userServices.js
@@ -5,8 +5,8 @@
  */
 
 angular.module('userServices', []).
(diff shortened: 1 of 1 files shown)


Gaelan:
What if we, instead of diffing after the fact, just keep track of our actions?

PR #100, opened 2014-01-19
fix(calculateDelta): don't detrain stats on ripening To-Dos

### LINKED ISSUE #2247: Automatic Allocation
I checked the "automatic allocation box" and I now got the third time in a row 1 point to strength, although 7 of my 8 daylies are mental ones. 

Shouldn´t my character get a point to int attribute?

SabreCat:
Previously, on cron, the reddening of To-Dos would subtract from the stat training for task-based auto-allocation. This had a tendency to cause users with lots of active tasks to get backward stat allocations on levelup. Now the "down" direction for To-Dos is not counted for attribute training. (Missed Dailies will still detrain.)

Fixes https://github.com/HabitRPG/habitrpg/issues/2247

### DIFF
--- a/dist/habitrpg-shared.js
+++ b/dist/habitrpg-shared.js
@@ -12046,7 +12046,7 @@ var process=require("__browserify_process");(function() {
                 }
               }
               if (task.type !== 'reward') {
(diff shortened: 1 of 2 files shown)


lefnire:
:+1: , let's remember this commit when we add "overdue to-dos hurt you" down the line

PR #116, opened 2014-01-26
adding mechanic to count previously owned pets

deilann:
changes hatched pets from 0 to -1

### DIFF
--- a/script/index.coffee
+++ b/script/index.coffee
@@ -550,11 +550,12 @@ api.wrap = (user, main=true) ->
         return cb?({code:404, message:":pet not found in user.items.pets"}) unless userPets[pet]
         return cb?({code:404, message:":food not found in user.items.food"}) unless user.items.food?[food.key]
         return cb?({code:401, message:"Can't feed this pet."}) if content.specialPets[pet]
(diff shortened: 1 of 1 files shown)


deilann:
Oh, right, and this also changes it so you can't feed pets if you own the requisite mount

lefnire:
Elegant, @deilann. Elegant.

deilann:
And this, children, is why you should never sleep.

lefnire:
[Here's what](http://gyazo.com/657d3b8dd15daf475a68542508e2a3a2) it looks like btw. Wont' see it until next push, unfortunately the new deployments will probably be pretty staggered since the `HabitRPG/develop`=>`lefnire/habitrpg-private` merge (added step for premium accounts) adds a bunch of overhead. I hope to peg down a better workflow sooner than later

deilann:
I can't tell if that's slightly too light, or it's my blistering headache.
What does 15% look like?

lefnire:
[15](http://gyazo.com/f539eb276663c41996baaeba02b57a40) [25](http://gyazo.com/fd153905254edc5ba794c899d7a4fedb)

deilann:
15 looks good to me, but part of color deficiency is things sometimes being
further "dimmed" than they are, especially later in the day. So I'd get
another pair of eyes on it.

PR #117, opened 2014-01-26
changing dequip to make sense for all classes

deilann:
Previously, all classes dequipped off hand equipment by selecting "No Shield" which makes no sense for folks like rogues, who don't use shields.

### DIFF
--- a/script/content.coffee
+++ b/script/content.coffee
@@ -156,7 +156,7 @@ gear =
   shield:
     base:
       0: text: "No Off-Hand Equipment", notes:'No shield or second weapon.', value:0
-      //changed because this is what shows up for all classes, including those without shields
+      #changed because this is what shows up for all classes, including those without shields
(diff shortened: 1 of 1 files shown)


colegleason:
Can you check the CI build failure?

deilann:
I can't figure out why it's saying there's a syntax error. It's quite
literally the same syntax, just with different text. I'll look at it later.

paglias:
I think it's the comment, in coffee they are written with a starting `#`

deilann:
...wow, I'm smart. This is what happens when I work in too many languages
at once and forget to switch my IDE.

On Sun, Jan 26, 2014 at 11:41 AM, Matteo Pagliazzi <notifications@github.com

deilann:
should be fix'd now

colegleason:
contrib.lvl++, grazi

PR #122, opened 2014-01-30
i18n(interpolation): add string interpolation to strings in jade files

marcelosizer:
Goes with: [habitrpg #2661](https://github.com/HabitRPG/habitrpg/pull/2661)

(@cc paglias)


paglias:
@lefnire I'll manually merge this and then the one on habitrpg, don't push immediately because this time I'll test everything is ok before pushing to github

PR #125, opened 2014-02-01
Changed some skills to use a hyperbolic formula

brandonjreid:
There have been problems with the formulas for Pickpocket #2093, Brutal Smash #2469, and Searing Brightness #2081.  I modified these formulas based on Aubec's excel sheet on the [Trello card](https://trello.com/c/L1H6LiZ3/110-character-stats-attributes-the-class-system).

Pickpocket is using his formula exactly with the assumption that task values < 0 should still be counted as 1.  This should make sure that the skill never gives you an obscene amount of gold.  I set the max to 75, but that can easily be changed.

For Brutal Smash and Searing Brightness I used values that would make the formula return a similar value as the linear formula using a stat of 200.  These skills have had problems with not showing any changes even after multiple casts.  This formula should make the tasks more effective at lower levels and taper off on higher levels.

Brutal Smashes max effectiveness raises the task value by 2.5 while Searing Brightness raises the value by 1.5.  That seems appropriate given that Searing Brightness affects all tasks, while Brutal Smash affects only 1 and can also crit.

### DIFF
--- a/script/content.coffee
+++ b/script/content.coffee
@@ -325,7 +325,7 @@ api.spells =
       target: 'task'
       notes: "You savagely hit a single task with all of your might, beating it into submission. The task's redness decreases."
       cast: (user, target) ->
-        target.value += user._statsComputed.str * .01 * user.fns.crit('per')
+        target.value += 2.5 * (user._statsComputed.str / (user._statsComputed.str + 50))
         user.party.quest.progress.up += Math.ceil(user._statsComputed.str * .2) if user.party.quest.key
     defensiveStance:
       text: 'Defensive Stance'
@@ -365,7 +365,8 @@ api.spells =
       target: 'task'
       notes: "Your nimble fingers run through the task's pockets and find some treasures for yourself. You gain an increased gold bonus on the task, higher yet the 'fatter' (bluer) your task."
       cast: (user, target) ->
-        user.stats.gp += (if target.value < 0 then 1 else target.value+1) + user._statsComputed.per * .075
+        bonus = (if target.value < 0 then 1 else target.value+2) * user._statsComputed.per
+        user.stats.gp += 75 * (bonus / (bonus + 800))
     backStab:
       text: 'Backstab'
       mana: 15
@@ -419,7 +420,7 @@ api.spells =
       cast: (user, target) ->
         _.each user.tasks, (target) ->
           return if target.type is 'reward'
-          target.value += user._statsComputed.int * .006
+          target.value += 1.5 * (user._statsComputed.int / (user._statsComputed.int + 40))
     protectAura:
       text: 'Protective Aura'
       mana: 30
--- a/script/content.coffee
+++ b/script/content.coffee
@@ -325,7 +325,7 @@ api.spells =
       target: 'task'
       notes: "You savagely hit a single task with all of your might, beating it into submission. The task's redness decreases."
       cast: (user, target) ->
-        target.value += 2.5 * (user._statsComputed.str / (user._statsComputed.str + 50))
+        target.value += 2.5 * (user._statsComputed.str / (user._statsComputed.str + 50)) * user.fns.crit('per')
         user.party.quest.progress.up += Math.ceil(user._statsComputed.str * .2) if user.party.quest.key
     defensiveStance:
       text: 'Defensive Stance'
--- a/script/content.coffee
+++ b/script/content.coffee
@@ -365,8 +365,8 @@ api.spells =
       target: 'task'
       notes: "Your nimble fingers run through the task's pockets and find some treasures for yourself. You gain an increased gold bonus on the task, higher yet the 'fatter' (bluer) your task."
       cast: (user, target) ->
-        bonus = (if target.value < 0 then 1 else target.value+2) * user._statsComputed.per
-        user.stats.gp += 75 * (bonus / (bonus + 800))
+        bonus = (if target.value < 0 then 1 else target.value+2) + (user._statsComputed.per * 0.5)
+        user.stats.gp += 25 * (bonus / (bonus + 75))
     backStab:
       text: 'Backstab'
       mana: 15


SabreCat:
Happy to see someone working on this!

For Pickpocket, what are some comparative numbers vs. what we see now? Personally I'd love to see the output fall by a factor of 10 (spells should be an extra bit of oomph, not totally eclipse actually getting things done), but that may be too heavy a nerf for the Rogue lovers out there. :P

What about Burst of Flames? It's got the same basic issues as Pickpocket, aye?

brandonjreid:
I took a closer look at the original spreadsheet, and the formula for the linear equation was wrong.  The hyperbolic equation was based off of the data from the linear one (trying to get similar numbers, but having the curve trail off to avoid obscene amounts of gp).  In summary, this would actually just make the problem a lot worse.  So, I re-worked the numbers.  The basic equation is still the same:  a \* (bonus / (bonus + b)), but I set a to 25, b to 75, and bonus =  value+2 + (PER \* 0.5).

I started putting in some example output, but its difficult since its based on both PER and the task's Value.  Instead here are some graphs for different task values (1, 5, 10, and 20) and a PER range of 1 - 200.  This would set the maximum gp you could get from pickpocket to 25 (unless a crit is possible from pickpocket).  Let me know what you think.  The numbers could easily be adjusted further.

The Searing Brightness and Brutal Smash skills shouldn't have this problem since I didn't base them off the original spreadsheet (I only used the same basic formula).  I'll also start looking into Burst of Flames.

v1
v5
v10
v20

SabreCat:
Those look like phenomenal improvements to me! :+1:

deilann:
omg <3

brandonjreid:
I made the change mentioned in my last comment.  Do all of the skills need their math to be reworked?  If so, do we want to do them separately as problems arise, or do we want to do it all in one go?  If reworking the math is really needed I can go though each one and find numbers that work well, although I would need some direction on whether each skill is too effective or not effective enough.

SabreCat:
I think it'll suffice for the moment to address the egregious offenders (Pickpocket, Burst of Flames too effective; Brutal Smash & Searing Brightness not effective enough). I haven't noticed any grumbles about other skills that seemed legit--like, some people think the healing spells don't heal enough, but they feel about right to me.

lefnire:
Here's the [simulation output](https://gist.github.com/lefnire/8262636/raw) if anyone's interested. Really glad someone's jumping into this, thanks @brandonjreid! UUID for reward?

brandonjreid:
@lefnire No problem!  Glad I could help.  Now I need to look for my next issue to work on.  Unless you have any suggestions?

UUID: af58e0ad-3b96-49da-b03e-a5176a50c355

SabreCat:
@brandonjreid I'd suggest putting Burst of Flames through the same paces as discussed above, though I know as-is BoF has its defenders.

marcelosizer:
@SabreCat hahaha, yes it does! =p

BoF has already been nerfed once when it was released as it was dealing twice the damage I believe. A party member who plays as a healer said that he's feeling useless compared to mages, so I wouldn't object on nerfing BoF even though I would prefer buffing healers instead =p.

brandonjreid:
I looked at Burst of Flames.  It already uses a hyperbolic model, but the numbers might need some adjustment.  It increases the task's value more than Searing Brightness or Brutal Smash (even after the buff I gave those skills).  So, we probably need to either decrease the effectiveness of Burst of Flames a bit or further increase in effectiveness of Searing Brightness and Brutal Smash.

SabreCat:
Similar to my thoughts on Pickpocket, I don't think that ability usage should yield more payoff than accomplishing your real-life tasks. A casting of Burst of Flames should be equivalent perhaps to checking off a To-Do, not, like, four of them.

marcelosizer:
Do either of SB or BS do damage to bosses? Their outcome seems rather different then BoF's.

SabreCat:
Brutal Smash is supposed to deal boss damage, but Backstab isn't. Which doesn't make RPG sense, but was done that way so Rogues don't steamroll battle quests the way they do collection quests.

marcelosizer:
@SabreCat I tend to do very few to-dos and I keep a large list of dark red ones so when I mark one I get huge EXP and damage, way more then BoF. Last week I leveled up twice on a single day from two dark red to-dos! And dealt massive damage to the boss.

marcelosizer:
How much damage does an average fighter do without skills?

brandonjreid:
SB decreases the redness of all tasks, but the most it can increase the task's value is 1.5.  BS only affects one task, but can increase the value by 2.5 and can crit.  BoF increases the value by up to 4 and also gives you experience of up to 75 and can crit (affects both value and exp).

I would say we change BoF to increase the value up to 2 and give experience up to 40.

marcelosizer:
That's the point, mage's skills **should** be better then fighter skills as fighters can damage through STR while mages can't. That's why I wanted to see a comparative of fighter vs mage damage, as a mage shall deal much more damage through skills then a fighter while dealing much less through normal atacks. The problem is that - as it has been noted many times on this discussion - the objective of the game is to complete your tasks, so it should deal more damage then magic skills. 

It seems the issue is bigger then buffing or nerfing skills. Maybe make the primary attribute of each class count towards dealing damage through tasks?

deilann:
I think the answer here is that STR doesn't seem to be greatly affecting
damage -- it increases it by .4% per point, which doesn't seem to put it
anywhere close to a few BoF over the course of a day. Maybe the answer is
to nerf BoF slightly, but also slightly increase the amount of damage STR
points give to boss damage.

lorian:
Imo, BoF badly needs to be nerfed (again); I was actually just looking at the code right now.

I think the main problems are: 1) boss damage scales too well with stats: at max int of 250 and ~max task value of 35, it does nearly 40 damage for 10 mp, 2) max mp is WAY too high with int: I currently can have nearly 450 max mp, which means I can do 1500 boss damage at once (after maybe a week of saving mp Edit: I just did the math, and it takes exactly a week of 10 yellow to-dos a day to hit my mp cap), and 3) mp cost for BoF is too low for how ridiculously awesome it is (not even counting boss damage; 10mp for 75xp? sheesh).

My party members have actually requested I stop casting BoF during boss fights, because they find it understandably demoralizing to have me one-shot the boss, while they do negligible damage. I could, from what I understand, trivially solo Vice right now. This is no good.

My suggestions are 1) change BoF to be hyperbolic, with the maximum cap being no more than the damage of a few (1-3) red to-dos, 2) Make max mp (and possibly mp gained/day and /task) similarly hyperbolic, with a much lower cap (I'm thinking maybe 100mp, which would be two casts of the highest-cost spell), 3) I would understand if it's part of the game design to always have a 10mp spell, so you can cast at least one spell per day, but in that case, BoF needs to give less xp. From what I can tell, even at low-level stats, it still gives ~40xp, which feels way too high to me.

lorian:
Heretical but genuine question: Would it be so bad to just have no spells that directly damage bosses? You damage bosses by doing your tasks, and maybe party buffs and healing can help you survive longer, but otherwise it's all on you and your productivity. Honestly, I think I'd prefer it like that.

deilann:
Problem with that @lorian is then mages have no tooth. They can't really
deal damage at all. I'd like spells not to affect, but boss damage would
need to be reassessed in general then.

deilann:
Mages would deal far less damage, as STR increases the damage done by .4%
per point.

lorian:
...Shoot, I accidentally deleted my comment, sorry.

Anyway, since everyone gets half/their full level in every stat, that doesn't seem to matter much at all. The mages in my party consistently do the most damage, well over that of similar-level rogues (no one is a warrior in my party, so perhaps I'm missing the awesomeness of a warrior; but I doubt it). 

And do note: because mages level faster, they will tend to be higher level than the rest of their partymates, which means they'll have higher strength than might be expected.

marcelosizer:
In regards to nerfing BoF it seems that it's decided, but in regards to max MP I'm completely against it. That's the pro of being a mage: huge mana pool. Mages cast spells, if you don't want to cast spells be a fighter, fighters swing swords. Also I'm against capping the use of spells as that's what MP is for.

@lorian the problem is that people play the game in different ways. My party barely deals damage to bosses if not by spells. I focus much more on habits and dailies then to-dos and I'm probably not the only one, and since to-dos deal way more damage then the rest BoF is not that OP to me.

lorian:
@Sinza- If max mp is nerfed, I think BoF has to be less nerfed, because max possible burst damage will be lower; if max mp is not nerfed, I think BoF needs to be much more heavily nerfed, so I can't solo Vice.

First off, fighters cast spells as well; the paradigm of Habit is that everyone has spells, not just mages. And it's certainly not that I don't want to cast spells, that's silly: I just want legitimate challenges, that encourage me to be productive. I was hoping Vice would be that; clearly it won't.

I think there are definitely ways to "nerf" my playstyle (I do about 10 to-dos and 30 dailies a day, and thus get around 65mp/day on the low end) without making yours unplayable, but I actually thought nerfing max mp was going to be the easiest, least controversial way. Do you have any suggestions for how to keep someone like me from soloing Vice once a week without any risk, or is that acceptable behavior as far as Habit is concerned?

Alternately, yall could start putting out actually hard bosses; I was thinking of suggesting something with 5-10k health. That would take me at least a couple weeks to solo!

deilann:
@lorian Epic Tier bosses are a'comin'.

lorian:
@deilann Woot! I look forward to being stomped on.

marcelosizer:
@lorian I know fighters have spells, the point is their spells should damage WAY less then mage's spells while their dailies and to-dos should deal WAY more damage. If you nerf max MP there's no reason to add points to INT.

I seldom do any To-dos, so I only get MP through cron. If you complete all your dailies and get no damage you are doing exactly what you should be doing: being productive. So you should reap the rewards, be them spells or custom rewards through gold. If you can accomplish that much you are truly an Archmage and killing Vice in a week is something that you can do with your eyes closed. ;)

marcelosizer:
Also, I believe there are cards on Trello to implement different difficulty levels and that's probably the way to balance your game.

lorian:
@Sinza- "If you nerf max MP there's no reason to add points to INT."
I put points in int for the increased xp gain; there's also the increased mp gain at cron and for each to-do, which can be turned into even more xp gain via BoF. Max mp seems like the least useful thing int can do; all it lets you do is store up BoF casts (since the other spells are useful only once a day). And I'm not saying there should be NO max mp increase for mages; just that it shouldn't be 450 at 200 int. If the max max mp was, say, 200, Vice would be out of one-shot range.

I do look forward to increased difficulty, but I doubt that's coming any time in the near future. Changing the way BoF scales, however, is something that can get done in a day or two.

marcelosizer:
@lorian you're right, MP could even be fixed like HP.

PR #127, opened 2014-02-05
Don't retry 400 level errors

colegleason:


paglias:
But if the user checked off a task or made any other change they will be on the client but not on the server so when the user check off the task again what will happen? If the server is reachable we may want to fetch the user from the server and apply it. Or there's something I missed?

colegleason:
Yeah possibly. 400 errors would indicate the client is possibly out of sync
with the server.

colegleason:
Alright, I've been thinking about this and I'm going to propose something a bit crazy (because it would take a lot of work):

The client is always getting out of sync with the server. We are always asking people to refresh. Is there a model that we can to refactor to where we don't show ops on client side unless the op completed with a success code?

I know this would mean the app wouldn't really work if the user was offline. But I feel like our current model just causes more grief than needed.

@lefnire

lefnire:
I'm all for removing offline from the web app if it increases stability. to be honest, offline was just a perk that came along with sharing code with mobile, but I figure not too many desktop users use it. I will say, offline use on mobile is a must have, so we'll want to keep that in mind with userServices (we don't have to share the file, we can support offline for mobile and not for desktop). 

Anyway, I'm gonna address the rest of that question tomorrow. I'm fried

lefnire:
oh wait, are you suggesting using standard routes instead of batch-update then? the only thing is we'll not want to deviate too far from shared code b/w mobile & web, for the sake of maintainability. Hmm... let me sleep on this...

colegleason:
No we don't have to deviate there. I just mean that the user will send an
op, it hits some bug/error (404), the client and server are now not synced.

I suppose they will sync up when the error comes back though. I don't know.

On Fri, Feb 7, 2014 at 1:14 AM, Tyler Renelle notifications@github.comwrote:

> oh wait, are you suggesting using standard routes instead of batch-update
> then? the only thing is we'll not want to deviate too far from shared code
> b/w mobile & web, for the sake of maintainability. Hmm... let me sleep on
> this...
> 
> ## 
> 
> Reply to this email directly or view it on GitHubhttps://github.com/HabitRPG/habitrpg-shared/pull/127#issuecomment-34409141
> .

## 

## Cole Gleason

Email: cg@colegleason.com
Website: colegleason.com

paglias:
We can store the previous client status and if an error occurs revert to it
or in case of an error (but not when the server is unavalaible) send the
whole user object back from the server to the client

colegleason:
Closed in favor of habitrpg/habitrpg#2775

PR #129, opened 2014-02-05
Replace hard-coded strings for window.env.t on js files

marcelosizer:
Goes with: [habitrpg #2762](https://github.com/HabitRPG/habitrpg/pull/2762)

(@cc paglias)


paglias:
@Sinza- I'm merging now, I think we may want to remove the "_commentchallengesctrljs" and similar "titles", what d yoou think?

We already have file division, unless you find them useful.

marcelosizer:
You can remove if you want, I just wasn't sure where to put them =p

marcelosizer:
I mean, I'll remove them. I'll upload the other languages and remove these comments from english, ok @paglias ?

paglias:
Wait a minute, I can remove them. I'm also removing app.json (we'll load all the files in a locale automatically) and changing some things. i'll post here when I'm done.

marcelosizer:
ok

paglias:
I'm done, removed app.json, kept the comments because they might be useful. Just tell the translators that they can skip them.

Now that we don't have anymore `app.json` if you create empty folders they won't be tracked in git until you add a file to them. You can add an empty `.gitkeep` file (the name can be anything but usually it's `.gitkeep`).

About `locales-old`, what to do with that? We can delete these folders or try to reuse some strings but I don't like that approach since a lot has changed in the meantime.

marcelosizer:
I was thinking on creating the folders with the english files so translators can use them as base. I think we can delete the locales-old folder.

paglias:
Copying the files, if something that isn't translated changes in the
english folder that is not yet translated in the other locale it would use
the old untranslated english version from the locale folder... But ok, I
think we can manually update everything when necessary, let's go ahead this
way and since you're there feel free to delete locales-old

2014-02-06 Marcelo Sizer notifications@github.com:

> I was thinking on creating the folders with the english files so
> translators can use them as base. I think we can delete the locales-old
> folder.
> 
> ## 
> 
> Reply to this email directly or view it on GitHubhttps://github.com/HabitRPG/habitrpg-shared/pull/129#issuecomment-34341213
> .

## 

marcelosizer:
Ok

PR #130, opened 2014-02-06
Stealth balance

lorian:
Made Stealth scale # of dailies avoided, based on user's # of total dailies. Note that it currently counts all dailies, including those not due today; I could make it count just active ones, but it'd be messy (unless someone knows an easy way).

I made it hit 100% at 100 total per, which is either a level 50 after a perfect day, with max points in Per, or a level 100 after an imperfect day, with half their stat points in Per. Feel free to balance as necesary!

I also fixed passive_active_attrs.coffee, which was not clearing buffs between tests, and added some dummy dailies (15 of them) so it can properly test the new Stealth.


lefnire:
aw wonderful! With tests and everything, you're wonderful @lorian! Got your ID from previous commit, contrib.lvl coming

PR #136, opened 2014-02-08
Update stats after casting spell

brandonjreid:
I could be wrong, but I think this is all we need to do to fix those issues of not leveling up after casting spells (Burst of Flames, and Backstab). Typically updateStats is called with a temporary stats object, but since all of the spells update user.stats we should just be able to call it with user.stats as the parameter.

It causes some of the statements in updateStats to not do anything (e.g. user.stats.gp = stats.gp), but I don't see it causing any issues.  Instead of decrementing the exp from a temporary stats object, it just decrements it from user.stats before increasing the level by 1.  

Fixes issue #2143 and #2598.


brandonjreid:
To clarify my previous comment:
Adding 'user.fns.updateStats user.stats' after a spell is used causes a player to be leveled up correctly after using BoF or Backstab.  I didn't see any problems with the change and looking at the code, I can't see any scenarios where this would cause an issue.

lorian:
So, when I tried that, it only caused levelup on the backend, and wasn't visible until after a browser refresh, rather than automatically (like when a task levels you up). Did you find a way around that?

brandonjreid:
Well, the habitrpg-shared code is both in node_modules/ and public/bower_components.  I made sure the change was in public/bower_components and recompiled it with grunt.  I don't know for sure if thats how it should be done, but it worked for me.

colegleason:
Yeah, I often find the issue there is that public/bower_components/habitrp-shared doesn't match node_modules/habitrpg-shared. Also check console though to make sure it isn't an error, @lorian

lorian:
Hmm. I have bower_components/habitrpg-shared as a symlink to node_modules/habitrpg_shared, so that should definitely not be a problem. I'll have to test again later when I have time.

brandonjreid:
I set my machine up to use a symlink (which is a whole lot nicer).  I noticed I had to run grunt in node_modules/habitrpg_shared before refreshing the page for the changes to show up on the browser end.

brandonjreid:
Is there any update on this?

raymondcascella:
I'm just getting started with contributing. Literally just got my local up and running a few moments ago =-D, so bear with me.

But, that means I did a fresh install to my local and I think things were moved around?? I don't see your code from 0753795 in the content.coffee file. It looks like it should be in the habitrpg-shared.js file, around line 10735? Something like user.fns.updateStats(user.stats);

paglias:
@raymondcascella you don't see it because it isn't merged yet

@brandonjreid sorry for the delay, I'll give it a look in the following days, thanks!

paglias:
@lefnire can we merge it?

lefnire:
Sorry so late, I'm looking into this now. This should work just fine, because it modifies `user`, and not `members` (or other). This is why I lagged on this so much. Here's what I mean:

On the server the members aren't loaded completely - only their attributes which can be affected by spells (currently `stats achievements items.special`, [here](https://github.com/HabitRPG/habitrpg/blob/37be3032d8216f8327f17239b2ce66be7a87596b/src/controllers/user.js#L341)). `updateStats` modifies additional attributes (eg `flags`, `preferences`). If those attributes are modified, mongoose will flag them to be saved and their values will become corrupt. To better explain, scenario:
- Load `member1.{stats achievements items.special}`
- `updateStats(member1)`. `member1.flags.itemsEnabled` becomes `true`, but because that value wasn't loaded, `member1.flags = undefined` to allow this set operation
- `member1.save()` and now `member1.flags = {itemsEnabled:true}`, and all other flags lost

**Anyway**, again all this is useless info regarding this PR, since this PR modifies the fully-loaded `user` model. I just wanted to dump some info for posterity in case someone tries a similar PR against updating members. I'm gonna test this PR a bit and get back to ya

lefnire:
After leveling up on a BoF, I started getting errors. The only one I caught was `can't read "language" of undefined`. I'll check into this again in the near future, gotta head out

brandonjreid:
@lefnire The i18n code has been added in since I made this change.  The updateStats function now has the req parameter (which I don't pass in because it didn't exist), so thats where the issue is coming from.  I'll have to look at this again and figure out the right way to solve this.

brandonjreid:
With all the changes since this has been written, I'll just close this pull request and create a new one when I have the bug fixed.

PR #138, opened 2014-02-09
Fixes beastmaster to not count gryphons

lorian:
Splits up drop and quest pets, and only counts the former (on the stable screen, and for beastmaster achievement). Mount screen now doesn't include rare mounts in its count. Also added additional unit tests for pet count, mount count, and beastmaster achievement.

This was done by creating content.dropEggs and content.questEggs, in addition to the original content.eggs. If anyone has a cute way of combining dropEggs and questEggs into content.eggs automatically, that would be extra-lovely; at the moment, the lists are written out manually, and will all have to be updated with new pets (assuming a more elegant solution isn't implemented by the time the next pet quest is released). Note that there is probably excess code in both content.dropEggs and content.questEggs, but I wasn't sure what was safe to remove.

This should resolve issue #2676 (https://github.com/HabitRPG/habitrpg/issues/2676), by separating quest pets both visually and achievement-wise. Note: please double-check that I didn't screw up spacing between lines of pets in stable.jade; my local install is showing wide spacing between each row, and I can't find any obvious cause.

(See pull request #2801 on habitrpg for rest of code)


lorian:
Oh, also: I added a couple strings to locales/en/pets.json, but I have less than no idea how to propagate that to the rest of the locales.

tests/algos.mocha.coffee
```diff
@@ -1,3 +1,8 @@
+### Must install: ###
+### sudo npm install -g mocha ###
+### npm install sinon ###
+### Run with mocha algos.mocha.coffee --compilers coffee:coffee-script ###
```
colegleason: You can replace this line with `npm test`

tests/algos.mocha.coffee
```diff
@@ -1,3 +1,8 @@
+### Must install: ###
+### sudo npm install -g mocha ###
+### npm install sinon ###
```
colegleason: This line can be `npm install --dev`. Sinon will be installed automatically then since it's listed in package.json.

script/content.coffee
```diff
@@ -527,7 +561,11 @@ api.hatchingPotions =
 _.each api.hatchingPotions, (pot,key) ->
   _.defaults pot, {key, value: 2, notes: "Pour this on an egg, and it will hatch as a #{pot.text} pet."}
 
-api.pets = _.transform api.eggs, (m, egg) ->
+api.pets = _.transform api.dropEggs, (m, egg) ->
+  _.defaults m, _.transform api.hatchingPotions, (m2, pot) ->
+    m2[egg.key + "-" + pot.key] = true
+
+api.questPets = _.transform api.questEggs, (m, egg) ->
```
colegleason, line 550: Quick questions, @lefnire, will all quest pets be based on their egg type? Are there plans to have special hatching potions?

colegleason:
Damn, you beat me to this one ;-). Great work!

For combining the two, try `api.eggs = _.assign(_.cloneDeep(api.dropEggs), api.questEggs)`.

lorian:
Good to know about the easier install strategy; I worked out what I did based on googling error messages, so it's not surprising it's suboptimal! I figured I'd at least leave a breadcrumb trail behind me for the next poor soul who tries to write tests.

And wow, I would not have figured that combination method. I tried a couple underscore methods, but none of them seemed to like objects, and by that point, I was getting really sick of eggs :P

colegleason:
Sorry about the messy install instructions. We actually need to change package.json's definition of `npm test` to use the correct `mocha`.

One thing to note is that we use LoDash, not underscore.js (tripped me up too). But yeah, it takes a while to get used to thinking that way.

tests/algos.mocha.coffee
```diff
@@ -1,3 +1,8 @@
+### Must install: ###
+### sudo npm install -g mocha ###
```
colegleason: Just tested: you can remove this line too. `npm test` will use the mocha installed by `npm install --dev` which is better.

lorian:
There we go; not only did I steal your idea, but I've stolen all your improvements too!

colegleason:
Awesome stuff!

script/content.coffee
SabreCat, line 550: Based on the specs out there for pixel artists, I don't think there are currently any plans for special hatching potions. That'd be a mad can of worms...

colegleason:
contrib.lvl++ btw.

PR #148, opened 2014-02-11
Added attribute associations for stock items

tmas:
Taking stairs is strength-based, reading is intelligence-based, etc.

Partially implements HabitRPG/habitrpg#2220


lefnire:
a fair point. though, i don't think users will ever see this - you unlock classes at 10, and task attrs only show if you have classes unlocked + auto-allocate=taskBased. By 10 they'll have long deleted these tasks. But it's useful for documentation sake if nothing else, thanks!

PR #149, opened 2014-02-12
Optimizations

paglias:


paglias:
It goes with https://github.com/HabitRPG/habitrpg/pull/2827

PR #151, opened 2014-02-16
Updated food items

baconsaur:
I'm not sure if I'm doing this right but I put the updated food items here!


lefnire:
Great! You're totally doing it right, wonderful job. Looking at the timestamp, looks like you did this in record time right after our G+

PR #155, opened 2014-02-18
Overhaul drop chance to a percentage multiplier system

SabreCat:
Turning this into a PR because it's easier to follow conversations over multiple commits that way.

I could use input from users interested in how often drops should happen, and statistics experts:
- Is the base chance OK? (about 2.5% for a "value 0" task). It can be adjusted two ways: by changing the divisor so each unit of task value is worth more or less, or by tweaking the subtracted constant to fine-tune the offset.
- Are the values of the factors OK? These are adjusted via the divisors, as you can hopefully see from the examples.
- Is this list of factors OK? Are there any that don't belong, or others we should include?
- PER/10 adds to daily drop cap: yes, no?

I could also use an assist in running a comparative drop chance simulation; I haven't done any tests of the sort yet.


SabreCat:
I tweaked something in the algos test to see if that clears up the Travis build failing here. @lefnire / @paglias / @colegleason, is there a way to rerun Travis without adding another commit?

paglias:
@SabreCat can't you run tests locally?

lefnire:
@SabreCat right, run `npm test`

SabreCat:
OK, will do! Not going to fix the "build failed" flag on this PR, though.

SabreCat:
Yay, I did the merge commit right (no junk file edits)! Will see if the build goes through OK, then merge and see how the user base takes it.

lefnire:
Long live sabe! I'll tell you when I push

PR #159, opened 2014-02-20
fixed space issues: contrib modal/challenge list

thepeopleseason:
Spacing issues in contrib modal and challenge list. 
Undocumented in [habitrpg issue #2582](https://github.com/HabitRPG/habitrpg/issues/2582), but some mentioned in #2875 and #2776.


thepeopleseason:
Oh well. Turns out my other, closed pull request would be preferred because the translation mgmt software doesn't handle spaces in locale strings.

PR #164, opened 2014-02-28
fix(strength): Remove threat reduction, increase crit chance

### LINKED ISSUE #2464: Rethink STR effects
A few users have noted that STR is the only "double-edged sword" attribute: reducing task redness might insulate you from future damage, but it also reduces the reward for getting your act together. It's the only attribute that has an inherent drawback.

This, however, seems to be mostly a philosophical objection, since the effect of even an off-the-charts STR score is unnoticeable when it comes to task threat. I just tested with a 1st-level chump, a 50th-level moderately-equipped Rogue, and a 100th-level Warrior with Golden Sword and Helm and every point invested in STR... and could not tell the difference when clicking a +/- Habit or Daily.

If it has barely any effect, and people don't much like the idea of it having a major effect in this way, maybe we should change it up?

There's not a lot left for STR to do, admittedly. But critical hits are cool, and a well-rounded sort of thing for the "I'll just play the default" users... what if it gave a bit better _chance_ for critical hits (like many people think it currently does), as well as change the _magnitude_ of the crits? The effect should be slow, not more than a tenth of a percentage chance per STR and probably even less than that, but it'd make it all-around desirable and if not immediately noticeable, at least palpable over time.

SabreCat:
Tweaks the STR attribute's algorithmic effects. (1) Task "redness" no longer declines faster with high STR. (2) The 3% chance of a critical hit is multiplied by a percentage factor equal to STR. (3) The value of STR in dealing damage to bosses is increased by 25% (.5% instead of .4%).

Fixes [HabitRPG issue 2464](https://github.com/HabitRPG/habitrpg/issues/2464).


SabreCat:
Don't merge yet, this is buggy. Testing now along with further drop chance tweaks.

SabreCat:
Argh! Junk commits. Closing and redoing with a cherry-pick scheme.

PR #165, opened 2014-02-28
Changed "Wizard" text to "Mage" in quests.json + Hedgehog Typo

DanielTheBard:
@lefnire - Fixed the first PR. Not touching the other one because I think @SabreCat wants to move the saddle to the food section in the store. So my PR fix wouldn't be needed.


lefnire:
Thanks duder!

PR #167, opened 2014-03-04
Some modifications for french locale.

Seldszar:


deilann:
I believe all of this is being done through Transifex instead.

Seldszar:
Oh, I didn't seen, sorry ^^'

PR #168, opened 2014-03-04
Translate content and api messages

paglias:
I've finally started working on adding translations for content (items, quests...) and api messages.

The idea is to move all the strings from plain js strings to functions that returns strings, for example:

```
0: text: "No Weapon", notes:'No Weapon.', value:0
```

would become:

```
0: 
  text: 
    -> lang.noWeaponText, 
  notes: 
    -> lang.noWeaponNotes, value:0
```

It's the only idea that come to my mind considering that keeping plain strings the lang object would have to be avalaible at compilation time and could not be changed, so it couldn't be used on the server.

The lang object could be a first level property of the habitrpg-shared module and we might allow it to be passed as a parameter to the function that return the translated string so that it wouldn't be necessary to change `habitrpg-shared.lang` every time, for example on the server where multiple requests requiring different languages might be executed at the same moment.

Also, I've checked that when we require a module in a part of the server and then requires it in another file it should share the same object.

Anyway the biggest point is the transformation from strings to functions, are you ok with that?
- [ ] **Update mobile app to work with the new code!**

@lefnire @colegleason @deilann @SabreCat


paglias:
I've started moving the strings to the new format, let me know if there's any problem with that.

paglias:
I need some help here, a lot of items in the gear under `weapon` have duplicates under `shield` (same text & notes), can we make so they share the same text & notes or they need to stay different and separated? @SabreCat

SabreCat:
That's for Rogues, I think, who wield duplicate Weapons in their "shield"
slot. They can safely share strings.

On Thu, Mar 6, 2014 at 12:37 PM, Matteo Pagliazzi
notifications@github.comwrote:

> I need some help here, a lot of items in the gear under weapon have
> duplicates under shield (same text & notes), can we make so they share
> the same text & notes or they need to stay different and separated?
> @SabreCat https://github.com/SabreCat
> 
> ## 
> 
> Reply to this email directly or view it on GitHubhttps://github.com/HabitRPG/habitrpg-shared/pull/168#issuecomment-36920750
> .

## 

The Hall of Twisted Mirrors http://sabrecat.wordpress.com/

deilann:
Yeah, I'm pretty sure that's what's up.

paglias:
:+1:

paglias:
@lefnire do you think it will be possible to upload this branch (not now, in a few days) somewhere so that translations can be tested?

The "techinical work" is almost done, I only need to move some messages to transifex but it will need some more work to make it so we can have grammatical correct translations because now it is organized on English which is quite different (and also simpler) than most languages.

PR #172, opened 2014-03-20
Aligned Nunchaku's transparent pixel with the character's hand.

PeterHatch:
Moved the Nunchaku down and in one 3x3 block, so the transparent pixel is aligned with the character's hand, like the rest of the weapons.

A bit concerned that it now goes too far down; perhaps more changes should be made?  But having the transparent pixel in the wrong place seems like a problem.


deilann:
Can you toss up a screen shot?

PeterHatch:
alterednunchaku

lefnire:
LGTM, UUID?

PeterHatch:
c8f8805d-d69e-452e-988f-9225d57d2d0d

deilann:
lvl++ :+1: 

Next one'll be tougher. ;)

PR #173, opened 2014-03-20
Fix #730: id property ignored when updating task

jbutz:
When calling the user.updateTask function if there was an id in the body it would override the id from the request URI.


lefnire:
UUID?

jbutz:
cb15097c-1743-4ce7-a769-55cf391d4cef

lefnire:
√

PR #177, opened 2014-03-23
Remove some unused images.

PeterHatch:
Removed broad_armor_special_tarataru.png, which is the same as broad_armor_special_2.png (and special_2 is the name of the slim and shop versions), and shop_head_healer_.png, which is the same as shop_head_healer_2.png.


deilann:
@lefnire Does removing unused images change everything on spritesmith?

PeterHatch:
Not lefnire, but I'm pretty sure it does.  Should I run grunt full and commit the changes to the files in dist?

deilann:
kk; if it changes everything on spritesmith that means that pretty much all
the images on the site are going to have to be fixed, as they use positions
on the sprite map.

Personally, I don't think removing two images is worth the amount of work
that entails. But yeah, run grunt full and let's see what happens.

PeterHatch:
Oh, okay, if it's work then it should probably, at best, wait until something else changes spritesmith and then be done at the same time.

(Is that work something that could be automated?)

deilann:
Wait, sorry, my brain isn't working this "morning." Spritesmith _is_ the
automation, so it should be fine. You've tested it, yeah? And the images
aren't all garbled?

Just gotta make sure that the unused items really aren't being used
anywhere and if they are, to fix that.

PeterHatch:
Yeah, I've tested and it seems fine.

deilann:
Awesome socks. Gonna leave this one for @lefnire as well.

lefnire:
Yeah, that's how spritesmith works. He just needs to run `grunt full` and everything gets fixed automatically. In fact, we usually accept non-grunted image updates since I'll typically run `grunt full` before pushing anyway. The whole system is automated, was a huge win in asset scalability (we _love_ spritesmith).

Cool beans, thanks @PeterHatch!

lefnire:
UUID? (@deilann, would you do the honors?)

deilann:
I've got his UUID. I'm thinking 4 small fixes for level 4? Did first one
and second one as levels 1 and 2, next two as 3.

lefnire:
:+1: exponential (ish) difficulty gain

PR #178, opened 2014-03-23
Gain the class bonus when using event equipment that belongs to your class.

PeterHatch:


deilann:
@lefnire this lgtm but I don't know if you want class equip bonus for special gear?

(Sorry Peter, I don't want to merge good things if I don't know if they're wanted. XD Hence me pinging Tyler.)

PeterHatch:
No problem!  I'm aware I could ask if things are wanted before doing them, but I enjoy doing things considerably more than asking.  And since I enjoy doing them and learn from it, I'm not really worried the effort would be wasted if it turns out they're not wanted.

deilann:
:) I'm _pretty_ sure this is wanted and I know there's a ticket about it.
And you're right-- it's more fun doing without asking.

lefnire:
I dig it! I think the addition of `specialClass` inadvertently allowed for fixing the old issue of special class gear not getting stat boost (per this PR), didn't even think of that! @SabreCat, I'll let you make the final judgement call since you had major thoughts about this - feel free to merge or close accordingly

SabreCat:
I like it the way it is, but not strongly enough to resist the will of the people on it!

lefnire:
A poll then? (or not worth the effort? I'm completely neutral, so y'all's call)

deilann:
It does confuse and upset some people, so idek.

SabreCat:
Green button clicked :p

PR #179, opened 2014-03-23
Made Clover-steel Helmet boost strength instead of perception; increased...

PeterHatch:
... cost of Hook Claws to match Ninja-to and Ski.


PeterHatch:
Also removed duplicate setting of specialClass on the spring shields.

deilann:
Personally, I wouldn't change the price, as the items have already come
out. However, the clover-steel helmet thing does need to be fixed as well
as the dupe setting. Can you split the commits so I can merge them
separately and see what the cofounders think about changing prices?

PeterHatch:
Not familiar enough with Git to know of a way to split the commit, so I just added a commit that undoes the change to the price - hopefully that works?

deilann:
Awesome, thanks. This is actually a much more simple solution.

deilann:
lvl++! Happy green!

I'm gonna arbitrarily say level 4 would be 4 or so of the smaller commits, but if any one who hands out more contrib levels than me disagrees, feel free to speak up. :P

PeterHatch:
Thanks!

lefnire:
This @PeterHatch character is on the rise!

PR #186, opened 2014-03-27
Aligned Crystal Sword to pixel grid, and added a transparent pixel to its hilt.

PeterHatch:


PeterHatch:
Here's the current version:
old crystal sword

And here's how it looks with the changes:
new crystal sword

PeterHatch:
(The difference is moving it left 1 pixel, to properly align with the 3x3 pixel block grid, and adding one transparent pixel where the hand goes, as is normal for the weapons.)

PR #188, opened 2014-03-28
Moved Jean Chalard's Noble Tunic up one pixel to align with the pixel grid

PeterHatch:
Here's the old version:
old jean chalard

And the new version:
new jean chalard


deilann:
Welcome to level 5!

PeterHatch:
Thanks!

PR #189, opened 2014-03-28
Create spritesmith CSS in a single pass.

PeterHatch:
Not only does this make the build process faster, but dist/habitrpg-shared.css becomes significantly smaller because we are no longer duplicating the CSS for every sprite that didn't need a .customize-options version.


PeterHatch:
I should mention the way I tested this - I ran grunt full before making the change, and copied the results to dist/original.  Then after making the change and running grunt, I ran git diff --no-index dist/original/customizer.css dist/spritesmith.css, and the only changes were adding the non-customized versions of classes before the customized version (i.e., .beard_ before .customize-option.beard_).  Then I ran git diff --no-index dist/original/spritesmith.css dist/spritesmith.css, and the only changes were adding the customized versions of classes after the non-customized versions.

Since neither one had anything in it removed or changed, I'm quite confident it's just adding to each what was missing from the other, and that the single-pass file nicely contains all the needed output.

Gruntfile.js
```diff
@@ -11,47 +11,34 @@ module.exports = function(grunt) {
...
+          // For hair, skins, beards, etc. we want to output a '.customize-options.WHATEVER' class, which works as a
+          // 60x60 image pointing at the proper part of the 90x90 sprite.
+          // We set up the custom info here, and the template makes use of it.
           if (sprite.name.match(/hair|skin|beard|mustach|shirt|flower/) || sprite.name=='head_0') {
-            sprite.name = 'customize-option.' + sprite.name;
-            sprite.x = sprite.x + 25;
-            sprite.y = sprite.y + 15;
-            sprite.width = 60;
-            sprite.height = 60;
+            sprite.custom = {
```
paglias, line 28: this way it doesn't create the spritesmith only for the customize-option class and not for the default one?

lefnire:
This looks great, I tested locally and all the usual suspects I look for checkout just fine! @paglias, want me to wait for a response per your concern (comment)? Else i'll merge locally & push (due to merge conflicts)

paglias:
@lefnire if it works for me is ok!
Il 01/apr/2014 00:01 "Tyler Renelle" notifications@github.com ha scritto:

Gruntfile.js
PeterHatch, line 28: The basic strategy is to not change any attributes, so the default output is still there, but to instead only add the custom property (which probably should have a better name...), which the template is modified to output additional data from if it is there.

Does that make it clear?

paglias, line 28: Yes, thanks!
Il 01/apr/2014 19:05 "PeterHatch" notifications@github.com ha scritto:

PeterHatch:
Just a reminder that this is still here...

lefnire:
Checks out for me, good job @PeterHatch! @deilann honors?

deilann:
I think that be level 6!

PeterHatch:
Thanks!

PR #191, opened 2014-03-30
Separate index.coffee into several smaller files.

Gaelan:


deilann:
I'm not sure I like this. :c

I think I wouldn't know where to look for things now.

But maybe I'm just resistant to change.

lefnire:
Don't merge this just yet. `index.coffee` indeed used to be in multiple files based on function. It caused _many_ issues with circular dependencies re: `require()` and browserify on web and mobile. It's been a very long time, and I can't remember what the issues were specifically - but we spent quite some time trying to fix the issues before tossing in and merging it all to one file. Let me see if I can find the conversation on those issues, and see if you can verify they're no longer an issue with this PR

lefnire:
alas, this will be too hard to resolve conflicts at this point, and I'm not sure if we want to go this direction. Plus with https://github.com/HabitRPG/habitrpg/issues/3439 we'll likely be taking a different direction

PR #192, opened 2014-03-30
Lowered Ski Parka by one 3x3 pixel block.

PeterHatch:
The Ski Parka is currently misaligned with the shirt, which means there are black outlines from the shirt showing up when wearing the Parka.

Here's the old version, for broad body types:
skiparka-broad-old

And here's the lowered version:
skiparka-broad-new

This is the existing version on a slim build:
skiparka-slim-old

And the lowered slim version:
skiparka-slim-new


deilann:
At this point, I should just keep your UUID on a sticky note at my desk. ;)

PR #194, opened 2014-04-06
replace comma with dot before capitalization

iartarisi:
I don't want gems for this, thanks! :)


deilann:
I can still mark it in the hall though, if you give me your UUID. :)

PR #196, opened 2014-04-12
Removing level cap and free Rebirth (DONE AND TESTED)

deilann:
@lefnire this is what I have so far


deilann:
kk, so this is done in theory -- just needs to be tested. On this side, it looks like the build is failing because it doesn't have the necessary info from the habitrpg repo PR?

paglias:
@deilann there are some strange files commited here `script/index.coffee~`, `script/index.coffee.BACKUP.10831.coffee`...

deilann:
Yeah, I just realized that I accidentally did an add all. I'm fixing things.

PR #198, opened 2014-04-18
feat(archive_todo): add cron to archive todos

jbutz:
This is part 2 of habitrpg/habitrpg#3062. Part 1 MUST be merged first.
Part 1 pull request: habitrpg/habitrpg#3302

### DIFF
--- a/script/index.coffee
+++ b/script/index.coffee
@@ -1237,6 +1237,10 @@ api.wrap = (user, main=true) ->
           else
             task.value = task.value / 2
 
+      user.todos.forEach (task) -> # Archive todos
+        if !task.archived && task.completed && moment(now).subtract('days',3).isAfter(moment(task.dateCompleted))
+          task.archived = true
+
 
       # Finished tallying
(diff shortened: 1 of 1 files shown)


script/index.coffee
```diff
@@ -485,7 +485,9 @@ api.wrap = (user, main=true) ->
       # ------
 
       clearCompleted: (req, cb) ->
-        _.remove user.todos, (t)-> t.completed and !t.challenge?.id
+        user.todos = _.map user.todos, (t)->
+            t.archived = true if t.completed and !t.challenge?.id
+            t
```
lefnire, line 490: easier as `_.each user.todos, ((t)->t.archived=true if t.completed and !t.challenge?.id)`?

jbutz, line 490: I didn't think modifying it inside the function would update the array, though I may have been wrong

lefnire, line 490: `t` is object by reference, so updating it's attribs will update the source. But hey, if it aint broke!

PR #203, opened 2014-05-11
Removing level cap, free rebirth at 100

deilann:
Goes with https://github.com/HabitRPG/habitrpg/pull/3449

### DIFF
--- a/locales/en/rebirth.json
+++ b/locales/en/rebirth.json
@@ -21,5 +21,11 @@
     "rebirthOrb": "Used an Orb of Rebirth to start over after attaining Level",
     "rebirthPop": "Begin a new character at Level 1 while retaining achievements, collectibles, and tasks with history.",
     "rebirthName": "Orb of Rebirth",
-    "reborn": "Reborn, max level <%= reLevel %>"
+    "reborn": "Reborn, max level <%= reLevel %>",
+    "welcome100": "Welcome to Level 100!",
+    "intro100": "Now that you've hit level 100, you have the option to use the Orb of Rebirth for free at any time.",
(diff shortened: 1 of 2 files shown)


deilann:
@paglias What's up, you think? I'm still having i18n build fail related stuff -- but if you look at the diff, there shouldn't be anything related.

paglias:
@deilann tests are probebly broken with i18n but this build is running, I fixed a bug here which should make sure the correct branch is used https://github.com/HabitRPG/habitrpg/commit/f29eaab7190d07ab4c37da87644d430984b083b4, probably it was using habitrpg-shared#master

deilann:
seems legit :thumbsup:

lefnire:
Fixed up some tests, now they're failing exactly as you'd expect them to (level 100 checks from before). @deilann would you be so kind as to tackle those, update them to the new expectations? I'm gonna be working on some other tests updates

deilann:
I just pulled the level 100 test out entirely. It checks 99. It might be worth it at some point to check things like expected attribute points is 100 at level 105, but if the other level tests are checking out, there's no reason higher levels should be broken.

If you'd like a few higher level expected experience tests, I can toss them in there.

lefnire:
Pulling the 100 test there is legit (either that or adding a 105 as you say). The others are that `freeRebirth` is borking some post-100 tests. Exactly as you'd expect it too (which is actually how I knew this commit was √), but we'll want to account for it nonetheless. Run `npm test` to see, [here's a gist](https://gist.github.com/lefnire/69ed3dd0900d2051d05d)

deilann:
Yeah, I noticed that one. I'll fix it ASAP.

PR #204, opened 2014-05-11
if negative scrolls, allow purchase to work

deilann:
Should help https://github.com/HabitRPG/habitrpg/issues/2592 along with https://github.com/HabitRPG/habitrpg/pull/3451

### DIFF
--- a/script/index.coffee
+++ b/script/index.coffee
@@ -604,7 +604,7 @@ api.wrap = (user, main=true) ->
         item = content[type][key]
         return cb?({code:404,message:":key not found for Content.#{type}"},req) unless item
         return cb?({code:401, message: i18n.t('notEnoughGems', req.language)}) if user.balance < (item.value / 4)
-        user.items[type][key] = 0  unless user.items[type][key]
+        user.items[type][key] = 0  unless user.items[type][key] > 0
         user.items[type][key]++
         user.balance -= (item.value / 4)
         cb? null, _.pick(user,$w 'items balance')


script/index.coffee
```diff
@@ -604,7 +604,7 @@ api.wrap = (user, main=true) ->
         item = content[type][key]
         return cb?({code:404,message:":key not found for Content.#{type}"},req) unless item
         return cb?({code:401, message: i18n.t('notEnoughGems', req.language)}) if user.balance < (item.value / 4)
-        user.items[type][key] = 0  unless user.items[type][key]
+        user.items[type][key] = 0  unless user.items[type][key] > 0
```
lefnire, line 607: having trouble wrapping my mind what this will do for those issues. What this line said before was: "If `user.items[type][key]` hasn't been initialized on the user object, initialize it with 0".

deilann, line 607: Yes, and now it says if it hasn't been initialized or is negative, set it
to zero. user.items[type][key] needs more than a value -- it should always
have a positive or zero value.

lefnire, line 607: it will fail for "if it hasn't been initialized". 

``` coffeescript
a = {}
a.b > 0 # false
a.b < 0 # false
```

Maybe this? 

``` coffeescript
user.items[type][key] = 0  if !user.items[type][key]? or user.items[type][key] < 0
```

lefnire, line 607: Oh, I see it!

deilann, line 607: :) I thought this was prettier.

PR #208, opened 2014-05-12
Tour: updates for new toolbar design

Alys:
I'm REALLY not happy with the way I've worded these changes because they sound clunky and messy, but the original text is wrong now, to the extent that it would induce frustration rather than being helpful. It might be better to just delete the two sentences I changed. The day start settings might be a bit advanced for an introductory tour? The "Social" thing that I wrote would be better on its own Tour page.


lefnire:
Agile. Correct > non-clunky, :+1:

PR #211, opened 2014-05-15
fix incorrect text for Light Crystals, fix Plain Egg pluralisation, remove excess word

Alys:
questVice2CollectLightCrystal had the wrong text ("Vice's Shade") - I've changed it to "Light Crystals"

For questEggHuntCollectPlainEgg, I've pluralised the text (Plain Eggs) because that reads better in the messages it appears in (and plurals are used in questEvilSanta2CollectTracks and questEvilSanta2CollectBranches).

For questRatCompletion, I've removed "out" from this: "we let out down our guard"


deilann:
lgtm thnx!

deilann:
I really need to fix the tests. :x I have it as a priority tomorrow!

PR #216, opened 2014-05-17
fix bug where an un-checked to-do with checklist removes only part of reward

Alys:
When a to-do is checked off, the rewards are increased by the number of checklist items that have been completed. However as reported in https://github.com/HabitRPG/habitrpg/issues/3474, if the to-do is later un-checked, then the rewards subtracted are not being similarly increased by the number of checklist items completed. This fix will change that - the number of checklist items completed will have an equal effect on checking and unchecking a to-do.

This of course won't fix the problem of, upon unchecking, not being able to calculate the exact value of the rewards that had been given upon checking, so there will still be a discrepancy between checking and unchecking, but it will be much smaller than it is currently.

### DIFF
--- a/dist/habitrpg-shared.js
+++ b/dist/habitrpg-shared.js
@@ -13159,7 +13159,7 @@ api.wrap = function(user, main) {
                   return m + (i.completed ? 1 : 0);
                 }), 0) / task.checklist.length;
               }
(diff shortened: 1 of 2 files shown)


Alys:
Closing this because I didn't mean for two separate changes to get parcelled up together. The lesson is: always create a new branch for every change.

PR #220, opened 2014-05-18
show pet's name (Flying Pig) instead of key (FlyingPig) in feed notifications

Alys:
Currently, if you feed a flying pig, the notification says, for example, "FlyingPig" (no space between the words). This should change it to use the correct string from the appropriate locale file.


lefnire:
`egg` is a [string at this point](https://github.com/HabitRPG/habitrpg-shared/blob/73b8128af992dcdd6b088519f72e267d83483564/script/index.coffee#L570). @Alys, if you could - do a quick test on local before sending PRs

PR #221, opened 2014-05-22
fix #3513 - if contrib level is NaN, use 0

Alys:
Fixes (I hope) https://github.com/HabitRPG/habitrpg/issues/3513 (non-contributor user reported getting over 20 drops today, when drop cap should have been seven).

NOTE that private code will need to be edited too - ref https://github.com/HabitRPG/habitrpg-shared/commit/f49ed5436c60d6b135909a3c8714347d99499701


lefnire:
Woops, commented on the wrong thread. Anyway, I'm gonna merge this regardless since it will defend against that if it's ever the case.

PR #228, opened 2014-06-17
Fix for issues #3573 - Correction for strange pixel on gems.

robeden:
This corrects two pixels on each version of the gem:
- Missing dark grey pixel at the top
- Stray light grey pixel in the 11:00 position that should be transparent (difficult to see outside an editor)


deilann:
waiting on permissions verifications. :)

robeden:
@deilann - I sent the email. Let me know if you didn't get it.

deilann:
I'm not the one who gets the email, is the issue. So I need a verification
from @lemoness. Or possibly @veeeeeee?

veeeeeee:
Received and confirmed. (FTR, all three cofounders get the email.)

deilann:
Ah, good to know!

PR #230, opened 2014-06-21
innText update of groups.json 2nd try

wc8:
Trying again without incorporating username. This moves all the inn messages to "innText"
(Ideally, later this message would replace Daniel's Tavern welcome when users are resting in the inn.)


wc8:
step 1 of part 1 for https://github.com/HabitRPG/habitrpg/issues/3627

PR #233, opened 2014-06-22
Beastmaster: props -> congratulate (remove slang)

Alys:
"give this user props" is an idiomatic phrase that non-native-English speakers could have problems with, especially since "prop" is also a noun that refers to an object. 

A player in the Tavern today said, _"Yeah! Beast Master Quest Complete! :D What about this 'give this user props!', am I going to get some cool gear or sth? :D"_  I had to disappoint him with the sad truth. :)


SabreCat:
Mad congratulates!

PR #235, opened 2014-06-23
"No Back Accessory" for base back - fixes #3640

Alys:
Changes the text and notes for the base back gear from "No Helm" to "No Back Accessory".

### DIFF
--- a/locales/en/gear.json
+++ b/locales/en/gear.json
@@ -330,8 +330,8 @@
   "shieldSpecialSummerHealerText": "Shield of the Shallows",
   "shieldSpecialSummerHealerNotes": "No one will dare to attack the coral reef when faced with this shiny shield! Adds <%= con %> points to Constitution. Limited Edition 2014 Summer Gear.",
 
-  "backBase0Text": "No Helm",
-  "backBase0Notes": "No headgear.",
+  "backBase0Text": "No Back Accessory",
+  "backBase0Notes": "No Back Accessory.",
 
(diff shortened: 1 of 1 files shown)


Alys:
Fixes https://github.com/HabitRPG/habitrpg/issues/3640

SabreCat:
Hurrah!

Alys:
When changes like this are made, should something also be done on Transifex? I'm guessing that the updated English json file is automatically pushed to Transifex, which means that the new English text would appear when you viewed the English and other-language strings side-by-side (and so it would be obvious then that a new translation would be needed), but is there an automatic flag that tells the translators that they need to look at those strings again? Or do we need to manually tell translators what has changed?

SabreCat:
I've posed the same question before but didn't get a response. It'd be good to know!

lefnire:
not sure the answer meself, @paglias would know.

paglias:
@Alys @SabreCat When a string changes here on GH it needs to uploaded on Transifex where every string that has changed is set as if it wasn't translated ans needs to be translated once again. The ulpload and download op with Transifex is manual. I do it about 1-2 times a week or when I see a lot of changes.

Changes made for english here are automatically used on the site, for other languages I have to update Transifex. Doing now.

paglias:
@Alys @SabreCat I think translators get an email when something changes, at least I get it. https://github.com/HabitRPG/habitrpg-shared/commit/b9bd9c7970e821a4f556b3c0eb0af0f0c6c25b2a

PR #237, opened 2014-06-26
fix 2 spelling mistakes

Alys:
havock -> havoc
Habitca -> Habitica


lefnire:
Forgot about these strings, placeholders (you can tell I'm not much of a writer). Pray, wanna take a stab at replacing it altogether with something more epic / stylish? If you think it's fine as-is, I'll merge

Alys:
I'm BAD at epic. No imagination. I think it's okay though. It's sounding like the questDilatoryBossRageDescription is a description of the actual mechanics, and I think it's more important for that to be clear rather than stylish. Some of the players have had trouble understanding the epic aspects of the recent Bailey announcements, so less epic for this particular string is probably good.

lefnire:
:+1:

PR #238, opened 2014-06-27
Check if items are owned before equipping

phillipthelen:
This checks if the user actually owns the pet, mount or gear they are trying to equip. This also prevents users from setting things that do not exist. (before users could set the value for pets and mounts to any string. Not sure if this leaves possiblilities for xss attacks. but this fix also prevents that).


lefnire:
:+1:

PR #241, opened 2014-07-01
explain what clearCompleted does and label button more clearly

Alys:
I'm pretty sure some people don't realise that the "Clear Completed" button permanently deletes all of their completed to-dos. The current wording around the button could be thought to imply that "clear" means "archive immediately". This change makes it obvious what will happen.

See also https://github.com/HabitRPG/habitrpg/pull/3670

### DIFF
--- a/locales/en/tasks.json
+++ b/locales/en/tasks.json
@@ -1,6 +1,7 @@
 {
-    "clearCompleted": "Clear Completed",
+    "clearCompleted": "Delete Completed",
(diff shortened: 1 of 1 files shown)


Alys:
Here's an example of confusion that I've just found on [the wiki](http://habitrpg.wikia.com/wiki/To-Dos):

"... a button marked 'Clear Completed' that lets you archive your completed to-dos. Clearing completed to-dos can help improve the site performance of HabitRPG...
Completed to-dos are automatically cleared after three days (but you can still find them in your history)."

PR #256, opened 2014-07-29
Snowball and June Item Icons

lemoness:
Changes to a consistent 2x to avoid resizing issues. These three should not look noticeably different, but I am uploading them for consistency. The summer items will have the most noticeable changes.


lefnire:
now we're cooking!

PR #258, opened 2014-07-31
Candy Cane text boosts Perception not Strength

jobywalker:
The notes for "weaponSpecialCandycane" say that Strength is boosted, but it should be Perception


jobywalker:
UserID is: 7941e218-7b6e-42b7-989d-8b040c65aa17

PR #260, opened 2014-08-01
Fix User Reset

jobywalker:
5cd0039 added regression where reset on new user would fail because user.items.gear.owned is undefined.
This patch allows tests to get _much_ further than before.


lefnire:
tests! we need! UUID?

jobywalker:
My UserID is: 7941e218-7b6e-42b7-989d-8b040c65aa17

lefnire:
√, thanks again!

PR #262, opened 2014-08-01
Fix freeRebirth flag

jobywalker:
Coffee is assuming the negation (!) applies to the AND of the two tests
should be just !user.flags.freeRebirth = true.  Fixes 4 test errors.

UserID: 7941e218-7b6e-42b7-989d-8b040c65aa17

### DIFF
--- a/script/index.coffee
+++ b/script/index.coffee
@@ -1155,7 +1155,7 @@ api.wrap = (user, main=true) ->
           dialog: i18n.t('messageFoundQuest', {questText: content.quests.vice1.text(req.language)}, req.language)
       if !user.flags.rebirthEnabled and (user.stats.lvl >= 50 or user.achievements.ultimateGear or user.achievements.beastMaster)
         user.flags.rebirthEnabled = true
-      if !user.flags.freeRebirth = true and user.stats.lvl >= 100
+      if user.stats.lvl >= 100 and !user.flags.freeRebirth = true
         user.flags.freeRebirth = true
 
     ###


jobywalker:
This should also fix:  https://github.com/HabitRPG/habitrpg/issues/3350

"Can't buy Rebirth Orb"

deilann:
Actually, that's super old and unrelated. That was pre-level cap removal.

The flag has nothing to do with purchasing the orb.

jobywalker:
Ah ok.

script/index.coffee
```diff
@@ -1155,7 +1155,7 @@ api.wrap = (user, main=true) ->
           dialog: i18n.t('messageFoundQuest', {questText: content.quests.vice1.text(req.language)}, req.language)
       if !user.flags.rebirthEnabled and (user.stats.lvl >= 50 or user.achievements.ultimateGear or user.achievements.beastMaster)
         user.flags.rebirthEnabled = true
-      if !user.flags.freeRebirth = true and user.stats.lvl >= 100
+      if user.stats.lvl >= 100 and !user.flags.freeRebirth = true
```
lefnire, line 1158: dunno if there's an additional bug here or it was intended, but `and !user.flags.freeRebirth = true` assigns.. should be `and !user.flags.freeRebirth == true` ? (or even just `and !user.flags.freeRebirth`?

jobywalker, line 1158: Yup completely missed that. Will submit another patch.

deilann, line 1158: Totally overlooked that. Sorry Tyler.

lefnire, line 1158: aint no thang, the bug was there before anyway

PR #264, opened 2014-08-01
Fix assignment in freeRebirth test

jobywalker:
Fixes the assignment in the test to determine if should set the freeRebirth flag.


deilann:
Oh my goodness that passed test is so gorgeous.

PR #279, opened 2014-08-18
Removed generated files

efim:
Tracking correspondence of committed generated files with current committed sources is problematic.
And anyway unnecessary (as far as I can see) since these files are generated anyway on setup.


Alys:
Keep in mind that if these files are removed from the repo, then when you fetch and rebase from upstream/develop, you'll also have to manually regenerate these files to ensure they are using the latest code. At the moment, that second step isn't necessary because you're pulling down the latest files that have already been generated. This step won't be obvious to inexperienced contributors, or to any contributors who don't know about this PR, so if it's accepted, please add documentation to the appropriate wiki pages.

lefnire:
> And anyway unnecessary (as far as I can see) since these files are generated anyway on setup.

TMK there's nothing in the website's run script that kicks this off, so the files won't be generated on our server (Heroku). Am I mistaken? If not, could you look into adding a grunt task on the website for generating these?

efim:
I compared newly generated file in develop branch with the one that is stored in repo right now and they don't really match. I will check this again, but it wouldn't be surprising if they came out of sinc. And it would be surprising if *.js file is not generated anew on server with every update of *.coffee file. 

So I guess right choice would be to generate it on main server with every update. This is only way that guaranties that they will stay in sync.

lefnire:
Yeah, this PR will go out of sync with every push that modifies those files (most pushes will) - don't worry about it, once we (you?) get the server triggering the habitrpg-shared build steps in place, I'll manually merge this PR & handle the conflicts

lefnire:
right, we'll need a system for kicking of builds when run before simply deleting the dist files

PR #280, opened 2014-08-22
keep tasks visible while editing - fixes https://github.com/HabitRPG/habitrpg/issues/1015

Alys:
Allows a task to stay visible while being edited, even if the edit has changed an attribute that would normally cause the task to be hidden.

**Example**: You have tag filtering turned on. You edit a task to remove the tag from it. Without this PR, the task immediately disappears and you have to turn off tag filtering to find it and save your edit. With this PR, the task remains visible until you have finished editing. The tasks disappears only AFTER you have saved it.

The "beingEdited" class is not currently used by HabitRPG but it could be used by third-party browser extensions.

**Example**: Deilann's [Hide Completed Dailies](http://habitrpg.wikia.com/wiki/Hide_Completed_Dailies) extension causes dailies to be hidden when they are not scheduled for today. The extension can be modified to use the beingEdited class to keep the daily visible even while the user is unscheduling it for today. The class could also be used by HabitRPG itself in future if we create our own feature to hide unscheduled dailies.

### DIFF
--- a/dist/habitrpg-shared.js
+++ b/dist/habitrpg-shared.js
@@ -6853,23 +6853,20 @@ process.chdir = function (dir) {
 
 }).call(this,typeof self !== "undefined" ? self : typeof window !== "undefined" ? window : {})
 },{}],4:[function(require,module,exports){
-(function (global){
 //! moment.js
-//! version : 2.8.1
+//! version : 2.4.0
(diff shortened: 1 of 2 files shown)


Alys:
I'm pretty confident that this won't break anything and will be a big improvement to usability, so I'll merge it in a day unless anyone stops me. :)

PR #281, opened 2014-08-23
accompanies reorganization of mounts page, in PR #3906

quin-jay:
Accompanies the pull request at [https://github.com/HabitRPG/habitrpg/pull/3906](https://github.com/HabitRPG/habitrpg/pull/3906)

I updated the mounts page to be organized the same way as the pets page, and so accordingly I added mounts in a couple of files in habitrpg-shared. I consider this pull request and the one at habitrpg to be part of the same contribution because they go together, but for the sake of having my UUID handy on both posts, I'll include it here too.

Thank you~! <3

1534a0ee-2f69-4c2e-9659-0d719ac2ea68


Alys:
```
  for pet of content.questMounts
    count-- if mounts[mount]
```

Is "pet" correct there?

quin-jay:
Ooh, good catch! I'll try 'mount' instead

ETA: I'm suddenly having a lot of trouble with actually getting github to recognize that I changed a file, so I'm going to work on this more in the morning.

quin-jay:
...sooo, I learned a lesson today. That lesson was: don't push more changes unless you want them to show up under the same pull request. My understanding is that you can pick and choose which commits to merge? So um, ignore the one with the sprite files for this PR, it's not related at all :$

~~i thought i'd float that idea later, cuz i felt like the green background under the potions made some of the colors a little hard to distinguish~~

Alys:
Hi @lynnyliz. I've been testing your changes to the Mounts page on my local install and have run into a few problems.  I'm going to close this habitrpg-shared PR without merging it for a couple of reasons, although the same changes could be resubmitted later if you wanted to.
- There are whitespace errors that were preventing your *coffee file changes from being compiled. Whitespace consistency is very important in coffeescript. The code in our coffee files uses two spaces for each indentation level, and never uses tabs, but your api.countMounts change had a tab in it instead of four spaces (it's entirely possible that your text editor did this automatically - mine does that sometimes). Also the new evolve function didn't have the necessary indentation. If the files were compiling on your local install, then there's something a bit inconsistent there - perhaps your local install has been configured to ignore whitespace errors, if that's possible?? You might want to work out what allowed to the code to compile locally with whitespace errors and make it stop permitting that. :)
- After I fixed those compile errors, I found that the changes to api.countMounts didn't make any difference to the Mounts page - it still displayed the total from gen1 and quest mounts together. It would be nice to get that fixed, but I think we should get the Mounts page reorganisation finished first (it's very nearly ready!), and then the count can be fixed up in a later change.
- The image changes that were included in this PR should be done separately. This sort of glitch is easy to do when you're new to git - I did it at least twice when I was starting out. :) The trick is to always create a new branch for each new feature that you are working on. Having different branches lets you work on multiple different features simultaneously with none of the code changes interfering with each other. If you'd like more information about this, just ask!

quin-jay:
Okay, thank you for your help! I've had to reinstall Windows since submitting these commits, so I'll get a local install running again and see if I can work out those problems. Thanks for the advice, I had very little idea of what I was doing at the time with branches and all. I'll try to be more careful in the future

PR #288, opened 2014-08-26
ApiUrlService as preparation for custom endpoints

negue:
Since syncQueue always use "the old" API_URL there is no other (good) way to change the API_URL, so I created a Service for this. For the website itself won't break anything, since no one will change the API_URL.

After this is merged / or with the upcoming mobile pullrequest, it will be possible to change the mobile API_URL

### DIFF
--- a/script/userServices.js
+++ b/script/userServices.js
@@ -1,20 +1,19 @@
 'use strict';
 
-/**
- * Services that persists and retrieves user from localStorage.
- */
-
 angular.module('userServices', [])
   .service('ApiUrlService', ['API_URL', function(currentApiUrl){
     this.setApiUrl = function(newUrl){
       currentApiUrl = newUrl;
     };
     
-    this.getApiUrl = function(){
+    this.get = function(){
       return currentApiUrl;
     };
   }])
-
+  
+/**
+ * Services that persists and retrieves user from localStorage.
+ */
   .factory('User', ['$rootScope', '$http', '$location', '$window', 'STORAGE_USER_ID', 'STORAGE_SETTINGS_ID', 'MOBILE_APP', 'Notification', 'ApiUrlService',
     function($rootScope, $http, $location, $window, STORAGE_USER_ID, STORAGE_SETTINGS_ID, MOBILE_APP, Notification, ApiUrlService) {
       var authenticated = false;
@@ -68,10 +67,8 @@ angular.module('userServices', [])
 
         // Save the current filters
         var current_filters = user.filters;
-        
-        var API_URL = ApiUrlService.getApiUrl();
 
-        $http.post(API_URL + '/api/v2/user/batch-update', sent, {params: {data:+new Date, _v:user._v, siteVersion: $window.env && $window.env.siteVersion}})
+        $http.post(ApiUrlService.get() + '/api/v2/user/batch-update', sent, {params: {data:+new Date, _v:user._v, siteVersion: $window.env && $window.env.siteVersion}})
           .success(function (data, status, heacreatingders, config) {
             //make sure there are no pending actions to sync. If there are any it is not safe to apply model from server as we may overwrite user data.
             if (!queue.length) {
--- a/script/userServices.js
+++ b/script/userServices.js
@@ -4,9 +4,19 @@
  * Services that persists and retrieves user from localStorage.
  */
 
-angular.module('userServices', []).
-  factory('User', ['$rootScope', '$http', '$location', '$window', 'API_URL', 'STORAGE_USER_ID', 'STORAGE_SETTINGS_ID', 'MOBILE_APP', 'Notification',
-    function($rootScope, $http, $location, $window, API_URL, STORAGE_USER_ID, STORAGE_SETTINGS_ID, MOBILE_APP, Notification) {
+angular.module('userServices', [])
+  .service('ApiUrlService', ['API_URL', function(currentApiUrl){
+    this.setApiUrl = function(newUrl){
+      currentApiUrl = newUrl;
+    };
+    
+    this.getApiUrl = function(){
+      return currentApiUrl;
+    };
+  }])
+
+  .factory('User', ['$rootScope', '$http', '$location', '$window', 'STORAGE_USER_ID', 'STORAGE_SETTINGS_ID', 'MOBILE_APP', 'Notification', 'ApiUrlService',
+    function($rootScope, $http, $location, $window, STORAGE_USER_ID, STORAGE_SETTINGS_ID, MOBILE_APP, Notification, ApiUrlService) {
       var authenticated = false;
       var defaultSettings = {
         auth: { apiId: '', apiToken: ''},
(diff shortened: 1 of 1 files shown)


negue:
@lefnire @paglias @deilann What do you think? Is it good to merge? 

So that we could have custom API-URL on habitprg-mobile.

paglias:
@negue where is the old API_URL defined? Couldn't we move that to a angular's `value`?

paglias:
@negue I would just change https://github.com/HabitRPG/habitrpg/blob/f981223153532035d72c35d42412628145c40385/public/js/app.js and https://github.com/HabitRPG/habitrpg/blob/f981223153532035d72c35d42412628145c40385/public/js/app.js and https://github.com/HabitRPG/habitrpg/blob/5c309444e050160b7d7852d3f03508acde862395/public/js/static.js#L4 from `constant` to value which is configurable

negue:
The "old" API_URL on the website stays as is to "" - [app.js#L12](https://github.com/HabitRPG/habitrpg/blob/f981223153532035d72c35d42412628145c40385/public/js/app.js#L12)

If It would just work by changing constant to value I wouldn't had to add this ApiService :P

Here is the Problem / Workflow:
1. open the APP
2. app.js on mobile sets the URL to https://habitrpg.com/
3. all needed Services are loaded and created, so is the "User"-Factory - [userServices.js#L18](https://github.com/negue/habitrpg-shared/blob/apiUrlService/script/userServices.js#L18)
4. User-Factory has the "old" API-URL inside it, and uses this to this execute `syncQueue` 
5. Login with UserID, ApiKey and different API-URL
5a. AuthCtrl sets the NEW URL to angularJS Value [authCtrl.js](https://github.com/negue/habitrpg-mobile/blob/changeApiUrl/scripts/controllers/authCtrl.js#L78)
5b. AuthCtrl calls `$http.post(API_URL + '/api/v2/user/auth/local' ...`
5c. Auth was working, great now all is set.
6. Use syncQueue to get the logged user
6.a - Won't work since the "User"-Factory instance is still there and so is the syncQueue Function, which still has the old API_URL, so syncQueue tries to get the user on the old API_URL, which results in an error, and we can't reload the APP.

Thats why I had to this little Service, so that the syncQueue always have the current API_URL :)

paglias:
Right :) But I would suggest anyway to move away from API_URL in favor of a
service everywhere so that the url can be changed not only for user service
but also for groups, challenges...

2014-09-12 18:21 GMT+02:00 negue notifications@github.com:

> The "old" API_URL on the website stays as is to "" - app.js#L12
> https://github.com/HabitRPG/habitrpg/blob/f981223153532035d72c35d42412628145c40385/public/js/app.js#L12
> 
> If It would just work by changing constant to value I wouldn't had to add
> this ApiService :P
> 
> Here is the Problem / Workflow:
> 1. open the APP
> 2. app.js on mobile sets the URL to https://habitrpg.com/
> 3. all needed Services are loaded and created, so is the "User"-Factory -
> userServices.js#L18
> https://github.com/negue/habitrpg-shared/blob/apiUrlService/script/userServices.js#L18
> 4. User-Factory has the "old" API-URL inside it, and uses this to this
> execute syncQueue
> 5. Login with UserID, ApiKey and different API-URL
> 5a. AuthCtrl sets the NEW URL to angularJS Value authCtrl.js
> https://github.com/negue/habitrpg-mobile/blob/changeApiUrl/scripts/controllers/authCtrl.js#L78
> 5b. AuthCtrl calls $http.post(API_URL + '/api/v2/user/auth/local' ...
> 5c. Auth was working, great now all is set.
> 6. Use syncQueue to get the logged user
> 6.a - Won't work since the "User"-Factory instance is still there and so
> is the syncQueue Function, which still has the old API_URL, so syncQueue
> tries to get the user on the old API_URL, which results in an error, and we
> can't reload the APP.
> 
> Thats why I had to this little Service, so that the syncQueue always have
> the current API_URL :)
> 
> —
> Reply to this email directly or view it on GitHub
> https://github.com/HabitRPG/habitrpg-shared/pull/288#issuecomment-55426237
> .

## 

negue:
Once this is merged, every other Service could also get the ApiUrlService instead of the API_URL.

But currently it is only needed here, because once the APP gets his user-data it redirects the page and then every other page already gets the "new" API_URL, or do I missing something here?

Or should I change now all other Services to use my ApiUrlService? And then this would be merged?

paglias:
@negue i'm talking about all the other references to API_URL, not everything goes through userService https://github.com/HabitRPG/habitrpg/search?utf8=%E2%9C%93&q=api_url

negue:
The Website services wouldn't be affected by my change. 

No one would and can change the URL there.

paglias:
And what about groups on mobile?
Il 12/set/2014 18:47 "negue" notifications@github.com ha scritto:

negue:
OMG .... I didn't know that the mobile project, also uses the services from habitrpg ....

Now this makes more sense ...

Why aren't these services inside the shared-repo?

paglias:
Because we didn't move them to the shared repo (can be done) and some of
them like the challenges are still not in the mobile app (planned)
Il 12/set/2014 18:53 "negue" notifications@github.com ha scritto:

negue:
@paglias I am currently adding the ApiUrlService to the HabitRPG-Services, inside the controllers which also uses API_URL should also use ApiUrlService? Or just the services?

paglias:
If we do it let's change it everywhere, so that we only have 1 method to
access the api url. I would do a PR or ask @lefnire just to be sure he's ok
with this but I don't think there'll be problems

2014-09-15 21:02 GMT+02:00 negue notifications@github.com:

> @paglias https://github.com/paglias I am currently adding the
> ApiUrlService to the HabitRPG-Services, inside the controllers which also
> uses API_URL should also use ApiUrlService? Or just the services?
> 
> —
> Reply to this email directly or view it on GitHub
> https://github.com/HabitRPG/habitrpg-shared/pull/288#issuecomment-55641316
> .

## 

negue:
Alright the controllers are next

negue:
I've added many times this `js var API_URL = ApiUrlService.getApiUrl();` just to use it like this `$http.post(API_URL + "/api/v2/user/auth/local", data)` 

what about if I extend the ApiUrlService to export "API_URL + neededUrl" ?

Like this `$http.post(ApiUrlService.getApiUrl("/api/v2/user/auth/local"), data)` 

Or If you have a better function Name for this?

paglias:
Let's keep it simple with apiurlservice.get() + "url"
Il 15/set/2014 21:43 "negue" notifications@github.com ha scritto:

negue:
I just thought it could be shorter, but okay I'll just use apiUrlService.get() + "url"

lefnire:
i'm in and out a bit, will review when I can but if y'all think think this is fine go for it

negue:
@paglias If there is nothing more to do, I would like to merge this PullRequest and sometime later the mobile / website pullrequests.

Alys:
I suggest waiting a few days until we know the server is all good.

negue:
@paglias @lefnire @Alys So can I merge this? Or should I push back this pull-request a little more and wait?

lefnire:
Long as it all checks out, I won't have much time to test it personally in coming days

paglias:
@negue feel free to merge!

PR #295, opened 2014-09-02
report correct equipment/gear item lost on death (most of the time)

### LINKED ISSUE #2789: Wrong equipment item lost on death (server loses diff gear than client)
Today I died at rollover. A dialog indicated that my robe had broken, and I could see that I had my hat but not my robe. I repurchased the robe. Later, after reloading again, I was wearing my robe but my hat was gone (not in inventory, repurchasable). My HP, XP, and gold were unchanged so it doesn't look like I died a second time.

UUID: 9b508a2e-1d1b-44ee-a437-266583e756e3

Alys:
Fixes https://github.com/HabitRPG/habitrpg/issues/2789 most of the time. On my local install, when I die with this new code, I see the wrong item reported maybe 1 in every 20 times, as opposed to about 1 in 1 without this fix.

### DIFF
--- a/dist/habitrpg-shared.js
+++ b/dist/habitrpg-shared.js
@@ -6853,23 +6853,20 @@ process.chdir = function (dir) {
 
 }).call(this,typeof self !== "undefined" ? self : typeof window !== "undefined" ? window : {})
 },{}],4:[function(require,module,exports){
-(function (global){
 //! moment.js
-//! version : 2.8.1
+//! version : 2.4.0
 //! authors : Tim Wood, Iskren Chernev, Moment.js contributors
 //! license : MIT
 //! momentjs.com
 
 (function (undefined) {
+
     /************************************
         Constants
     ************************************/
 
     var moment,
-        VERSION = '2.8.1',
-        // the global-scope this is NOT the global object in Node.js
-        globalScope = typeof global !== 'undefined' ? global : this,
-        oldGlobalMoment,
+        VERSION = "2.4.0",
         round = Math.round,
         i,
 
@@ -6881,11 +6878,8 @@ process.chdir = function (dir) {
         SECOND = 5,
         MILLISECOND = 6,
 
-        // internal storage for locale config files
-        locales = {},
-
-        // extra moment internal properties (plugins register props here)
-        momentProperties = [],
+        // internal storage for language config files
+        languages = {},
 
         // check for nodeJS
         hasModule = (typeof module !== 'undefined' && module.exports),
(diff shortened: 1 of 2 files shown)


paglias:
@Alys what about the momentjs change in the commit?

Alys:
Do you mean something in dist/habitrpg-shared.js? I have no idea. It might have got pulled in from somewhere else. I didn't bother looking at what was in it since the entire file would be changed as soon as grunt was run on the server. I probably shouldn't have attached it to the commit.

paglias:
I don't think grunt is run on the server for habitrpg-shared yet. Anyway it
seems a downgrade in Momentmjs to 2.4 from 2.8
Il 03/set/2014 13:43 "Alice Harris" notifications@github.com ha scritto:

paglias:
@Alys 2.4 seems to be the right one https://github.com/HabitRPG/habitrpg-shared/blob/develop/package.json#L8 so don't worry

Alys:
Really??? Grunt's never run? Doesn't that cause major issues? When something as simple as having a different version of software to what's on production can change habitrpg-shared.js so much, couldn't that really mess with how the production site behaves?

paglias:
@Alys I think Tyler make sure to run grunt and commit the new version of habitrpg-shared.js each time he makes a deploy so there shouldn't be problems. It's necessary because bower doesn't have any concept of builds it must have the compiled version ready to work on the browser. But we could run grunt in habitrpg-shared directly from habitrpg postinstall script. I'll make a pr

Alys:
Oh, no, that deploy process is what I was referring to when I assumed that grunt was run on the server - I was meant the production server (or staging if you have a pre-prod server that you do the final compiles on). I didn't mean running grunt in habitrpg-shared itself. But if you want to set that up, I won't argue. :)

paglias:
Ah okok.

Anyway this change would be a little problematic and we would have to
change the mobile app build too so let's not make it, at the end is
just running grunt on the local machine before deploying

2014-09-03 13:57 GMT+02:00, Alice Harris notifications@github.com:

> Oh, no, that deploy process is what I was referring to when I assumed that
> grunt was run on the server - I was meant the production server (or staging
> if you have a pre-prod server that you do the final compiles on). I didn't
> mean running grunt in habitrpg-shared itself. But if you want to set that
> up, I won't argue. :)
> 
> ---
> 
> Reply to this email directly or view it on GitHub:
> https://github.com/HabitRPG/habitrpg-shared/pull/295#issuecomment-54285778

## 

PR #301, opened 2014-09-08
Update tasks.json

Lamecarlate:
Teeny tiny update: "vouloir supprimez" is not proper French :)


Alys:
Unfortunately, translations are not managed on GitHub, but in Transifex. There's [a few details here](https://github.com/HabitRPG/habitrpg-shared/blob/develop/locales/README.md) and it links off to more information.

PR #305, opened 2014-09-10
prevent off-class items and free items being lost on death

Alys:
Changes death so that you can lose only items that belong to your current class and items that belong to no class. See the [Trello death card](https://trello.com/c/k2MWBcxl/204-death-mechanic-v2) and https://github.com/HabitRPG/habitrpg/issues/2551

Also prevents free items being lost. Losing free items gives an unfair advantage to subscribers ("pay to win") because they have many more such items (Mystery Items) and so would have a greater chance of losing a free one. 

@lefnire or @paglias or anyone: I would love to know what's happening here with 'toObject' https://github.com/Alys/habitrpg-shared/blob/83b1804a555d11cc97634a7aa53da388311a2ad6/script/index.coffee#L401
It doesn't seem to be doing any damage, but ... what the heck??

### DIFF
--- a/dist/habitrpg-shared.js
+++ b/dist/habitrpg-shared.js
@@ -16,6 +16,8 @@ var process = module.exports = {};
 process.nextTick = (function () {
     var canSetImmediate = typeof window !== 'undefined'
     && window.setImmediate;
+    var canMutationObserver = typeof window !== 'undefined'
+    && window.MutationObserver;
     var canPost = typeof window !== 'undefined'
     && window.postMessage && window.addEventListener
     ;
@@ -24,8 +26,29 @@ process.nextTick = (function () {
         return function (f) { return window.setImmediate(f) };
     }
 
+    var queue = [];
+
+    if (canMutationObserver) {
+        var hiddenDiv = document.createElement("div");
+        var observer = new MutationObserver(function () {
+            var queueList = queue.slice();
+            queue.length = 0;
+            queueList.forEach(function (fn) {
+                fn();
+            });
+        });
+
+        observer.observe(hiddenDiv, { attributes: true });
+
+        return function nextTick(fn) {
+            if (!queue.length) {
+                hiddenDiv.setAttribute('yes', 'no');
+            }
+            queue.push(fn);
+        };
+    }
+
     if (canPost) {
-        var queue = [];
         window.addEventListener('message', function (ev) {
             var source = ev.source;
             if ((source === window || source === null) && ev.data === 'process-tick') {
@@ -65,7 +88,7 @@ process.emit = noop;
 
 process.binding = function (name) {
     throw new Error('process.binding is not supported');
-}
+};
 
 // TODO(shtylman)
 process.cwd = function () { return '/' };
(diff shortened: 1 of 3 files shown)


lemoness:
Excellent call about the free items being lost - that didn't even occur to me, but it's a very good point!

Alys:
Thanks Lemoness. I should have put a little bit more thought into it though. :) It was preventing new users from losing any gear at all on death because the only gear they had was the free sword. I'm changing it so that you can't lose any free gear except for the starter sword (weapon_warrior_0).

Alys:
The User Death "revives correctly" test is failing - it reports that for a new user, weapon_warrior_0 is NOT lost on death. However every test I've done on my local install since my code change a couple of hours ago indicates that the weapon IS lost. 

Interestingly, the Death "doesn't break unbreakables" test DOES succeed, which means that weapon_warrior_0 IS being lost in that test.

Does anyone have any ideas about what's going wrong during the "revives correctly" test or any advice about how I can troubleshoot this?

lefnire:
@Alys toObject:

In certain cases, you can't iterate over sub-properties on a Mongoose model directly. You can _access_ them directly (because Mongoose sets up proper setters/getters), but not iterate over them. So to get around that, you may have to call `_.each user.whatever.toObject()`. Instead of _skipping_ toObject, call it. See if that fixes the tests issue

lefnire:
To see why that's the case, setup a debugger on the server and set a breakpoint somewhere on the user. You'll see that every nested property is a complex object, with setters/getters for accessing that attribute at that point. So iterating it directly will cause some funky situations.

Alys:
@lefnire : Thanks very much for the toObject explanation. Improved code here: https://github.com/Alys/habitrpg-shared/blob/deathGear-Sep10/script/index.coffee#L399
But is there a better/neater way to do it?

The User Death "revives correctly" test is still failing, exactly as before, even though tests on my local install show that scenario working correctly every time. And as before, the Death "doesn't break unbreakables" test DOES succeed, which means that weapon_warrior_0 IS being lost in that test. I'm at a loss to explain why "revives correctly" is failing.

lefnire:
neater would be `a.toObject?() or a`. I'll check out the rest when I can

Alys:
I've improved the code as suggested [above](https://github.com/HabitRPG/habitrpg-shared/pull/305#issuecomment-55528155) by @lefnire: https://github.com/Alys/habitrpg-shared/commit/9350a3c715eb27a1b754e7bda4b882491f52420a

I've repeated my manual tests and it all still seems to work, including the case where a new character loses their training sword. And yet that automated test keeps failing. Can we assume that the test is somehow broken and make this live? I can test on a new user account as soon as this goes live.

We get a couple of users every week commenting sadly that they can't get back their off-class death-lost gear, and this PR would fix that.

Alys:
Unless there's any complaints in the next day, I'm going to merge this (fixing conflicts first, if needed). I still think that the automated test is wrong, and even if it's not, the consequences are easily tested for and not too dramatic.

Alys:
Rebased, tested manually, all works. Ignoring failed Travis tests because they lie!!! Probably. Will test in prod as soon as I see this go live.

PR #313, opened 2014-09-16
allows stats to change for users who used rebirth

Alys:
Prevents this problem from happening: https://github.com/HabitRPG/habitrpg/issues/4046

However it won't help the users that it has already happened to. :( I'm sorry i caused that.

@lefnire, are you able to fix their accounts when they post their UUIDs?

### DIFF
--- a/dist/habitrpg-shared.js
+++ b/dist/habitrpg-shared.js
@@ -12438,7 +12438,7 @@ module.exports = {
   strings: null,
   translations: {},
   t: function(stringName) {
-    var locale, string, stringNotFound, vars;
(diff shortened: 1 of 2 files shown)


Alys:
@lefnire, it looks like cron fixes the problem (it did on my local install just now, and for at least one production user), so you don't have to do anything for the two users who reported it in https://github.com/HabitRPG/habitrpg/issues/4046

But this PR should probably be merged and made live soon to prevent it happening to others. I'm sorry about the hassle. :(

lefnire:
Will push when home

PR #314, opened 2014-09-17
Fixed various typos in German translation

snipem:


thepeopleseason:
HabitRPG Translations are managed by Transifex, instead of directly in the locale files. You can find more information here: http://habitrpg.wikia.com/wiki/Guidance_for_Linguists

PR #315, opened 2014-09-17
Fix percent function to return minimum zero

sbreiler:
The percent function is used for ui-bars. When returning numbers below zero the bar will show up with 100% width, 'cause css don't know negative width. Issue #3886


Alys:
I've tested locally and it's all good. Merging now.

@sbreiler : Please tell me you [UUID (User ID)](https://habitrpg.com/#/options/settings/api) so I can give you a contributor tier.

PR #316, opened 2014-09-19
add coverage tools to habitrpg-shared

Feriority:
This should help with #174, though it only has coverage for the .coffee files, not the pure js.  As far as I can tell, though, that's all we're actually testing now.

I spent a while trying out some other tools for coverage, but ibrik couldn't process some of the files and karma-coverage seems to use ibrik to preprocess coffeescript.  Nothing seemed to cover both the js and the coffeescript well.

This is very easy to use - `npm run coverage` will generate a coverage.html file with nicely formatted coverage information.

I'm new to node, so please let me know if I'm doing something weird or violating best practices in some way.

### DIFF
--- a/.gitignore
+++ b/.gitignore
@@ -1,3 +1,4 @@
+coverage.html
 node_modules
 npm-debug.log
 .idea
--- a/package.json
+++ b/package.json
@@ -10,6 +10,7 @@
     "grunt-spritesmith": "~1.22.0"
   },
   "devDependencies": {
+    "coffee-coverage": "~0.4.2",
     "mocha": "*",
     "expect.js": "*",
     "sinon": "~1.8.2",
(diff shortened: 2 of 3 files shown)


crookedneighbor:
@Gaelan Does this fulfill what you wanted in #174 ?

crookedneighbor:
I made one modification, and that is to append `; open coverage.html` to the end of your `coverage` command in the package.json, that way it automatically opens the result in the web browser.

@Feriority Can you post your user id here? It can be found in [settings > api](https://habitrpg.com/#/options/settings/api)

Feriority:
f48afed8-3356-4fc0-9c6d-c90b06e7f335 - Thanks!

Alys:
Tier++. Thanks @Feriority!

The higher tiers require more effort than the lower ones, so if you make any small changes from now on, you might find that they won't grant you an extra tier immediately, but they will be noted as partial credit towards a tier. There's a bit of information about that at [Contributor Rewards](http://habitrpg.wikia.com/wiki/Contributor_Rewards) but do ask if you have questions.

PR #318, opened 2014-09-20
add markdown as angular filter

negue:


negue:
@lefnire @paglias It is good to merge? Since I would only add this filter.

Alys:
It's probably a really good idea to wait a few days until the server settles down before we make more changes to the site.

From the sounds of the latest updates, @lefnire is pretty busy at the moment and might not have time to advise.

paglias:
@negue sorry, what's the reason about that? I mean we already have the directive

negue:
You can't use a directive inside a "attribute binding".

paglias:
:+1: if it's not a problem let's wait for the rest to be solved as @Alys suggested before merging, for me it's ok to merge it when we can

negue:
Of course! First all bugs must be solved before we can add more features. :)

lefnire:
For my own, I don't want to be the bottleneck of PRs getting merged as I struggle with memleak + mongoose `defaul:{}` stuff. Up to y'all, sounds like the three of you agree on waiting - but I'd personally say "hey, keep 'em mergin! I'll be back"

negue:
I'll merge this tomorrow after work.

PR #319, opened 2014-09-20
Resizes pet-dragon-hydra.png

rowan-seda:
```
    Fix for issue #272. Also removes commented out reference in backer.css
```


crookedneighbor:
I understand decreasing the width, but can you tell me why you increased the height as well? Is that the standard height for the other rare pets?

If I'm not mistaken, the `grunt sprite` task will need to be run to compile this, but I expect that's something we'll want to do on this end after an image change like this gets merged in, since it'll result in modified css and png files.

crookedneighbor:
Oh, duh, I see now. You made it the same height as the other dragons.

Can you post your user id here? It can be found in [settings > api](https://habitrpg.com/#/options/settings/api)

rowan-seda:
Hi Blade,

Sorry I hadn't yet gotten to reviewing your earlier questions regarding grunt and the height. But I'm glad it became obvious.

My HabitRPG user ID is 08e01950-e209-4959-919c-5e60f229d703. 

Thanks,

Brandon Seda

Alys:
Tier++. Congratulations and thanks @bcseda!

PR #322, opened 2014-09-29
feat(pets): release pets

deilann:


deilann:
Goes with https://github.com/HabitRPG/habitrpg/pull/4093

PR #327, opened 2014-10-01
Remove the right amount of gold and exp when unchecking a task

brandonjreid:
There are a lot of code changes for this, but a good portion of it is splitting calculateDelta() into two functions.  The idea here is when a task is unchecked the gold and experience removed is different from the amount added.  When we uncheck a task, we need to change the task back to its original value (which involves approximating what that original value was), then calculate the delta off of the original value.  That delta can then be used for the exp and gold calculation.

I tested the approximation code for over 600,000 values between -47.27 and 21.27 and it always approximates the original value in 2 to 4 iterations, so this shouldn't affect anything speed wise.  I can upload the python code i wrote to test it, if you would like.

This doesn't fix any issues with finding an item on an accidental click, but at least the numbers come out the same now.

This fixes issue HabitRPG/habitrpg#964 along with countless other ones that I won't mention here :)

### DIFF
--- a/dist/habitrpg-shared.js
+++ b/dist/habitrpg-shared.js
@@ -14180,10 +14180,10 @@ api.wrap = function(user, main) {
           return nextDelta;
         };
         calculateReverseDelta = function() {
-          var calc, closeEnough, currVal, diff, testVal;
+          var calc, closeEnough, currVal, diff, nextDelta, testVal, _ref1;
           currVal = task.value < -47.27 ? -47.27 : task.value > 21.27 ? 21.27 : task.value;
           testVal = currVal + Math.pow(0.9747, currVal) * (direction === 'down' ? -1 : 1);
-          closeEnough = 0.0001;
+          closeEnough = 0.00001;
           while (true) {
             calc = testVal + Math.pow(0.9747, testVal);
             diff = currVal - calc;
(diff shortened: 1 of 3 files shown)


paglias:
@brandonjreid I know they are failing right now because we have to update a value but could you add a test for this? Great work, we got countless issues about that!

brandonjreid:
Yeah, I'll update any test cases and create some new ones.

brandonjreid:
I fixed a couple bugs with what I wrote, and added some test cases.
I also put in some code for checklist todos, so the gp and exp are reversed correctly for those.
Now this should also fix issue HabitRPG/habitrpg#2222

Alys:
https://github.com/HabitRPG/habitrpg/issues/2222 was fixed some time ago but was never closed. I've closed it now. It was not talking about the small differences in earnings that you get from ticking/unticking, but large differences because checklist items were not being counted on unticking.

brandonjreid:
@Alys Ah, I see.  Well, I accidentally reintroduced the bug with my first commit, but fixed it again with the second.

paglias:
@brandonjreid on it now, will resolve merge conflicts and then merge. This absolutely deserve a badge, can you give me your UUID?

brandonjreid:
@paglias af58e0ad-3b96-49da-b03e-a5176a50c355

paglias:
@brandonjreid awarded :) thanks a lot

PR #331, opened 2014-10-08
Changed the spell formulas for better balance.

brandonjreid:
Some of the spells are out of balance as seen in issue HabitRPG/habitrpg#3542. I decided to play around with these, and here's what I came up with.  Most likely we'll need people to play around with it and figure out if we need to tweak the numbers a bit more, but here's a first whack at it.

What I changed:
Wizard:

```
Burst of Flames:
    value: Slightly more effective for low level users and slightly less effective for 
        mid-high level users so its more in line with Brutal Smash and Searing Brightness.

    exp: Changed the calculation for code uniformity.  May be slightly less effective.

    boss damage: Reduced the damage from BoF, more so for the high level users.
        This spell in conjunction with a large mana pool was too powerful
```

Warrior:

```
Brutal Smash:
    boss damage: Increased the damage from Brutal Smash, especially for low-mid level users.
        Gave it diminishing returns, so the damage wouldn't get out of hand.
```

Rogue:

```
Pickpocket:
    code cleanup, but otherwise unchanged

Backstab:
    exp: Gave it diminishing returns.  Now it is in line with Burst of Flames.

    gp:  Gave it diminishing returns.  The gp gained is less than Pickpocket, since
        Pickpocket only awards gp, while Backstab awards gp, exp, and increases the task's value.
```

Healer:

```
Searing Brightness:
    value: Slightly increased effectiveness
```

Here are Some charts for a better visual of how its changed.  I did a couple 3d charts since the formula is based on the user's stat as well as the task's value. Keep in mind the scale is different in the 3D charts, so don't just look at the color.

image
bofold
bofnew

### DIFF
--- a/script/content.coffee
+++ b/script/content.coffee
@@ -408,6 +408,8 @@ api.gearTypes = gearTypes
 #
 diminishingReturns = (bonus, max, halfway=max/2) -> max*(bonus/(bonus+halfway))
 
+calculateBonus = (value, stat, crit=1, stat_scale=0.5) -> (if value < 0 then 1 else value+1) + (stat * stat_scale * crit)
+
 api.spells =
 
   wizard:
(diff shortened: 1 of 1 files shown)


Alys:
@brandonjreid At least some portions of this are being merged into the `rebalancing-2014-12` feature branch. If we end up using that code, it will count towards a contributor tier. Can you tell me your UUID please?

Alys:
I've tested this version of backStab and it seems good to me. I've merged it into the feature branch and marked backStab as done in our list of requirements. However if others want to test and provide feedback, they are certainly welcome to.

Other skills in this PR are also being considered - see https://github.com/HabitRPG/habitrpg/issues/3029#issuecomment-66400962 for details.

Alys:
I've tested this version of Brutal Smash and it's good too and has been merged.

Pickpocket was merged a few days ago but has not been tested yet (examination of the code made it seem okay).

We probably won't use this version of Burst of Flames because we had another PR that did exact percentages suggested by Ryan.

brandonjreid:
@Alys Sounds good.  Here is my UUID: af58e0ad-3b96-49da-b03e-a5176a50c355

crookedneighbor:
@Alys Did you notate the contributions from @brandonjreid ?

Can this PR be closed?

Alys:
No, please do not close this. I am still working with it. I will close it when I am finished. That will be soon.

Alys:
@brandonjreid I have awarded you a contributor tier for this. Thank you for your work!

PR #333, opened 2014-10-10
Dailies, To-Dos and Habits capitalization consistency

megrbui:
First contribution! For example: To-dos --> To-Dos

PR #334, opened 2014-10-10
Edited capitalization for consistency

beckie42:
Changed capitalization to be consistent


Alys:
Level will be awarded in https://github.com/HabitRPG/habitrpg-shared/pull/333 since this is basically the same change.

Alys:
@beckie42 : I'm really sorry! When I was merging pull requests I saw three capitalisation fixes at the same time and assumed they were all by the same person (but only two of them were). That was stupid of me.

Can you please tell me your UUID (Unique User ID) so that I can give you a contributor tier for this? You can find it under [Settings -> API](https://habitrpg.com/#/options/settings/api).

PR #335, opened 2014-10-10
Updated capitalization for consistency

megrbui:
e.g. dailies --> Dailies


Alys:
Level will be awarded in https://github.com/HabitRPG/habitrpg-shared/pull/333 since this is basically the same change.

PR #336, opened 2014-10-10
Fixed cron dayStart option.

brandonjreid:
This has been said before, but I think I fixed the cron issues with dayStart.

The issue comes when the last cron was after midnight and before dayStart.  Then when you attempt a cron again on the same day, but after dayStart, it doesn't compute the daysSince correctly.  I adjusted startOfDay to take dayStart into account.  So, when dayStart = 4 and lastCron = '2014-10-09 01:30:00' then startOfDay will return '2014-10-08 04:00:00' since technically our day doesn't start until 4am.  This makes the code for both daysSince and shouldDo simpler because we just alter which day we consider a certain timestamp to be at.

In daysSince I didn't use the diff 'true' parameter or use a smaller diff interval with Math.ceil as was suggested earlier because I compute the startDay of both yesterday and now.  So, both dates will be set to dayStart.  They might have a difference in date, but not in time, so the diff will always be a whole number.

I wrote a number of test cases for startDay and daysSince to verify they are correct. Some of the test cases in Cron->dailies->new day were incorrect.  For example: "due yesterday  due today  pre-dayStart  checked" was expecting cron not to run even though dayStart = 4, lastCron = '2014-10-09 03:00:00', and now = '2014-10-10 03:00:00'.  The incorrect tests were assuming that lastCron always happened after dayStart, even though lastCron was being set before dayStart.

I removed the test "calculates day differences with dayStart properly" because with the fixed cron the test would fail if you ran it between midnight and dayStart and pass other times.  The new tests I created should cover the same cases as this test, so I just removed it.

I also fixed the test "calculates the start of the day" for the time at 23:59:59.  The issue here was the date is in January, but the function wasn't given a timezone, so it defaults to now's timezone.  So, the test case fails if you run it while in daylight savings time because the timezone offset is different.  Passing in the timezone for January fixes this issue.

Should fix (crossing my fingers) issue HabitRPG/habitrpg#1057

### DIFF
--- a/script/index.coffee
+++ b/script/index.coffee
@@ -42,7 +42,10 @@ api.startOfWeek = api.startOfWeek = (options={}) ->
 
 api.startOfDay = (options={}) ->
   o = sanitizeOptions(options)
-  moment(o.now).startOf('day').add({hours:o.dayStart})
+  dayStart = moment(o.now).startOf('day').add({hours:o.dayStart})
+  if moment(o.now).hour() < o.dayStart
+    dayStart.subtract({days:1})
+  dayStart
 
 api.dayMapping = {0:'su',1:'m',2:'t',3:'w',4:'th',5:'f',6:'s'}
 
@@ -51,7 +54,7 @@ api.dayMapping = {0:'su',1:'m',2:'t',3:'w',4:'th',5:'f',6:'s'}
 ###
 api.daysSince = (yesterday, options = {}) ->
   o = sanitizeOptions options
-  Math.abs api.startOfDay(_.defaults {now:yesterday}, o).diff(o.now, 'days')
+  Math.abs api.startOfDay(_.defaults {now:yesterday}, o).diff(api.startOfDay(_.defaults {now:o.now}, o), 'days')
 
 ###
   Should the user do this taks on this date, given the task's repeat options and user.preferences.dayStart?
@@ -60,13 +63,7 @@ api.shouldDo = (day, repeat, options={}) ->
   return false unless repeat
   o = sanitizeOptions options
   selected = repeat[api.dayMapping[api.startOfDay(_.defaults {now:day}, o).day()]]
-  return selected unless moment(day).zone(o.timezoneOffset).isSame(o.now,'d')
-  if options.dayStart <= o.now.hour() # we're past the dayStart mark, is it due today?
-    return selected
-  else # we're not past dayStart mark, check if it was due "yesterday"
-    yesterday = moment(o.now).subtract({days:1}).day() # have to wrap o.now so as not to modify original
-    return repeat[api.dayMapping[yesterday]] # FIXME is this correct?? Do I need to do any timezone calcaulation here?
-
+  return selected
 
 ###
   ------------------------------------------------------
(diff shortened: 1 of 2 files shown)


paglias:
@brandonjreid I gave a look at the code and it seems a reasonable fix for me! I think now we'll have some testing time on beta or directly on prod

Great! 

_crossing fingers_

lefnire:
beta, for at least a week. I'll get back to y'all once I get it up there

deilann:
Crossing my fingers as well!

lefnire:
on beta now

lefnire:
been on beta for a while now, haven't heard any feedback. Likely the previous beta-testers of cron are scared, I wonder how we can get some eyes on this

brandonjreid:
@lefnire Maybe no news is good news? @lemoness posted test results on the issue (HabitRPG/habitrpg#1057) and didn't find any problems. It wouldn't hurt to have more eyes on it though.

The thing that assures me that it works is the fact that there were test cases that were written incorrectly, but they were passing.  Once I made the change to the code they were failing (and rightly so). I fixed the logic of those tests and they pass on the new code.  I also tested every case I could think of for daysSince and get the correct result for each one.

lemoness:
We could ask everyone in the Contrib guild to test it for a bit?

paglias:
I think we should probably add the option to sue beta from mobile. i couldn't test it because I use mostly mobile to check off tasks and see damage the morning

crookedneighbor:
Did we ever get any conclusive info from this? Is it worth rebasing and deploying to beta again?

lemoness:
I believe it still needs to be thoroughly tested.

crookedneighbor:
@lefnire and @paglias would it be possible to copy the data of one of the affected users, scrub out the personal data (replace task text, inbox messages, profile bio, etc with `SCRUBBED` or something), and see if the bug can easily be replicated locally? That'd make troubleshooting the cause of the cron bugs easier, right?

paglias:
@brandonjreid I've set up some debugging to discover errors on cron and it doesn't seem there are specific errors happening there, so it's probably a problem with the algorithm. We're going to test this pull request on beta in the next few days and see if it's working

paglias:
there have been some code changes in the meantime so I'm going to make some changes to make sure everything works as expected

paglias:
Merged in https://github.com/HabitRPG/habitrpg-shared/tree/paglias/cron-fix

crookedneighbor:
Looks like this was merged into the cron-fix branch, so I'm going to close this. Re-open if it's still relevant.

paglias:
Keep it open. We're still testing for a few days
Il 30/dic/2014 22:35 "Blade Barringer" notifications@github.com ha
scritto:

paglias:
merged 

@brandonjreid we're going to put this on production, I'll let you know how it goes

paglias:
@brandonjreid if you're still around there are some news:

We've tested your code in production for more than a day now and the only feedback we received is positive so far and we didn't notice any other issue, so it seems to be working. I'll wait for another few days then close HabitRPG/habitrpg#1057 and you'll be able to collect the bounties there.

The only problem we encountered is this https://github.com/HabitRPG/habitrpg/issues/1057#issuecomment-68928401 It's not related directly to your code but we would like to fix it anyway because it can be very annoying. I've posted [here](https://github.com/HabitRPG/habitrpg/issues/4488) an explanation of what I think is causing the issue along with a possible solution. What do you think?

Alys:
@brandonjreid I've awarded you a contributor tier for this. I'm sorry we didn't give it to you earlier.

brandonjreid:
@Alys No problem, thanks!

PR #341, opened 2014-10-17
Shortened title so popover glitch would be fixed

crookedneighbor:
When clicking on party member, a modal pops up. The "Class Equip Bonus" is wrapping to the next line. For some reason, the wrapping is causing the popover to flash briefly and then disappear. If you get your mouse in just the right spot, it'll appear correctly. Shortening the string fixes the problem.

skitch

I propose shortening the string to "Class Bonus". Hovering over it will show the full title "Class Equipment Bonus" and give an explanation of what it is, so I don't think it's a big deal to shorten it to "Class Bonus".

Thoughts?


crookedneighbor:
1) Just realized my screenshot is inaccurate. The part marked "Does not work" is from a screenshot I made after editting it using Chrome's dev tools, where I changed the string that it is pointing to.

2) Alternative to this could be to add some css specific stylings to the party member modal that cuts off the "Class Equip Bonus" text and adds an ellipses. Ie: `Class Equi...: number`

Alys:
Thanks @crookedneighbor!

PR #343, opened 2014-10-20
i18n uk: missed translations for settings added

holyspecter:


Alys:
I'm sorry but we can't merge this change because translations are managed on Transifex. If we merged this, it would be overwritten by later Transifex updates. 

If you'd like to help with the translations on Transifex, your help would be welcome. For details, please see the _README_FIRST.md file that's in the same directory as the file you changed, or the wiki's [Guidance for Linguists](http://habitrpg.wikia.com/wiki/Guidance_for_Linguists) page.

holyspecter:
@Alys, sadly I've look there after this PR was created. Ok, I'll use Transifex next time.

PR #344, opened 2014-10-21
Added @ symbol to arcosine

crookedneighbor:
Completion string was missing the @ symbol on the username


crookedneighbor:
At first I thought the "Completion: Completion: " was a convention for the questCompleted strings, but after looking at the other ones, that doesn't look like the case, so I removed that part.

Alys:
Thank you @crookedneighbor

crookedneighbor:
My pleasure.

PR #351, opened 2014-11-01
Added space before "target=" when added to href

spacetime:
Fixes bug where the first links in markdown boxes don't work
screen shot 2014-11-01 at 10 35 43 pm
screen shot 2014-11-01 at 10 35 21 pm
screen shot 2014-11-01 at 10 35 13 pm


lefnire:
good catch, thanks! UUID

PR #355, opened 2014-11-06
Adding URL sanitization to prevent security issues.

rhtgrg:
This prevents XSS. Note that this particular sanitization method will also make it impossible to make relative URL links in markdown. That can be added if needed.

### DIFF
--- a/script/directives.js
+++ b/script/directives.js
@@ -43,9 +43,12 @@
           .replace(/"/g, '&quot;')
           .replace(/'/g, '&#39;');
       };
+      var cleanUrl = function(dirtyUrl) {
+        return (/^https?:\/\//.test(dirtyUrl) ? dirtyUrl : '');
+      };
       if (cap[0].charAt(0) !== '!') {
         return '<a class="markdown-link" target="_blank" href="'
-          + escape(link.href)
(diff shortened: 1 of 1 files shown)


negue:
As far as I understand this RegEx it just test the url if it is https? Its been a long since I read some docs about xss but wouldn't it be still possible to use xss even with a https:// before?

rhtgrg:
No, I'm not talking about something like CSRF, I am talking about XSS of this format:

[hacked](javascript:alert%28document.cookie%29)

Of course, you can do just about anything once you can run JS, the above is really a non-issue compared to the other things you could do.

Anyway, the more secure the better of course. :)

crookedneighbor:
I just checked this out on my local branch, and I like the idea, but it seems to me like this kind of check should be done as the user submits the message, not after, and it should emit some kind of warning (message contains and invalid url). 

Does that make sense?

The way it is now, the message does get submitted, but when clicked on, it opens a link to the root of the site, in this case, it sends the user to the tasks page. I'd rather prevent the message from being posted at all, rather than have it post and send no where.

crookedneighbor:
Going to close this for now, if you'd like to work on a version of this that checks the url on submission, go ahead.

PR #361, opened 2014-11-18
Adjust golden knight (shouldn't get Jean Chalard's Tunic)

taldin:
Since you probably aren't supposed to get Jean Chalard's Tunic (the equipment reward from finishing the Recidivate Transformed quest) AND Mustelaine's Mashing Offhand, here's a quick fix to get rid of the line awarding the Tunic.   (Also apparently rebundles my unpicked up fix for rebuying the Main Hand Morningstar fix on my Develop branch.  Sorta oops.)

PR #363, opened 2014-11-19
Bug in the Mustaine's Morningstar Fix

taldin:
Had one too many 'or's in the fix (https://github.com/HabitRPG/habitrpg-shared/pull/361) when I pasted it over to the
repository.


taldin:
Never mind, I see you caught it already.  :)

PR #367, opened 2014-11-20
tavern-tier-rewards-update(en)

taldin:
Updated to show the correct gem awards for tiers in the tavern - they were still showing '2' across the board. (en corrected only, other localization files will need the same fix)   Tested and working on my local install.
image


Alys:
Blimey! How did we miss that??? I blame me. :(

I'll give this a bit of a test in a few hours, just to be sure, and then merge it.

Thanks taldin! Well spotted.

taldin:
<grins> I only spotted it because a friend asked me about the colors after I got bumped a tier this morning (...dunno that I feel I really earned it, since it was the 2nd tier Blacksmith bump for a minor bug fix...>.>) and I was saying it was one step closer to the Hydra pet.

Alys:
Tested, looks good.

@taldin Now that you're at tier 5, this isn't enough for a new tier, but I've made a note of it in your profile so it will count towards the next one.

taldin:
Sounds good.    Thankya!  Deleting the branch.

PR #366, opened 2014-11-20
Standardization of locales/en/gear.json

ShilohT:
A few days ago, Sulfirh mentioned in the Aspiring Legends guild that many of the Mystery Items were lacking Month and "Confers no benefit." notes.  I went to take a look at gear.json and saw a lot more things that were inconsistent. I welcome feedback and I hope I'm not stepping on any toes by doing this! 

**Changelog in the commit notes.**

**Some thoughts for discussion:**
Placement of "Confers no benefit." in the middle vs the end: I think of 'Confers no benefit.' as a stat effect tag, so it should be in the same place as that would be. I can sort of see how one might want it at the end instead though. Thoughts?
A few of the holiday accessories were using "Confers no stat bonus." instead.  I kind of like this phrasing better than "Confers no benefit.".  This is just a find/replace to change anytime.

Currently, all Mystery Items are 'Items' and all seasonal equipment is 'Gear'.  Is it okay to have both 'Items' and 'Gear'? If so, what makes something 'Gear'?  If it is that 'Gear' has a stat bonus, the seasonal accessories would then be 'Items'.  Or is 'Gear' default and only Mystery Items are 'Items'?  (I might have answered my own question there.)
If we're using both 'Item' and 'Gear': I changed the Convention accessories to say "Special Edition Convention Item" (instead of 'Armor'); perhaps they should be 'Gear' instead?

The Absurd Party outfit (175, 279) has no event tag right now.  Should it also be labelled 'Limited Edition 2013 Winter Gear'?  (Limited Edition 2014 New Year's Item/Gear?)
Should the special equipment in lines 65-74, 159-164, 271-276, 356-359 be tagged Quest, Contributor, etc?
The Rainbow Warrior outfit (178, 317) has a completely different description than anything else, so I wasn't comfortable with messing with it.

**And the beginnings of a style sheet:**
1. Description. 2. Stat Effect. 3. Event/Origin(if needed)

.2. Stat Effect:
No bonus:
Confers no benefit.
Single or all stat bonus:
Increases (Strength/Intelligence/Constitution/Perception/all attributes) by (<%= str %>/<%= int %>/<%= con %>/<%= per %>/<%= attrs %>).
Two stats, same bonus:
Increases (Strength/Intelligence/Constitution) and (Intelligence/Constitution/Perception) by <%= attrs %> each.
Two stats, different bonus
[...]

PR #372, opened 2014-11-27
Standardization of locales/en/gear.json

ShilohT:
This is the same as pull request #366 .  New pull request from new branch, because my develop branch got really messy. (and this fixes the merge conflict issue)

> A few days ago, Sulfirh mentioned in the Aspiring Legends guild that many of the Mystery Items were lacking Month and "Confers no benefit." notes. I went to take a look at gear.json and saw a lot more things that were inconsistent. I welcome feedback and I hope I'm not stepping on any toes by doing this!
> 
> **Changelog in the commit notes.**
> 
> **Some thoughts for discussion:**
> Placement of "Confers no benefit." in the middle vs the end: I think of 'Confers no benefit.' as a stat effect tag, so it should be in the same place as that would be. I can sort of see how one might want it at the end instead though. Thoughts?
> A few of the holiday accessories were using "Confers no stat bonus." instead. I kind of like this phrasing better than "Confers no benefit.". This is just a find/replace to change anytime.
> 
> Currently, all Mystery Items are 'Items' and all seasonal equipment is 'Gear'. Is it okay to have both 'Items' and 'Gear'? If so, what makes something 'Gear'? If it is that 'Gear' has a stat bonus, the seasonal accessories would then be 'Items'. Or is 'Gear' default and only Mystery Items are 'Items'? (I might have answered my own question there.)
> If we're using both 'Item' and 'Gear': I changed the Convention accessories to say "Special Edition Convention Item" (instead of 'Armor'); perhaps they should be 'Gear' instead?
> 
> The Absurd Party outfit (175, 279) has no event tag right now. Should it also be labelled 'Limited Edition 2013 Winter Gear'? (Limited Edition 2014 New Year's Item/Gear?)
> Should the special equipment in lines 65-74, 159-164, 271-276, 356-359 be tagged Quest, Contributor, etc?
> The Rainbow Warrior outfit (178, 317) has a completely different description than anything else, so I wasn't comfortable with messing with it.
> 
> **And the beginnings of a style sheet:**
> 1. Description. 2. Stat Effect. 3. Event/Origin(if needed)
> 
> .2. Stat Effect:
> No bonus:
> Confers no benefit.
> Single or all stat bonus:
> Increases (Strength/Intelligence/Constitution/Perception/all attributes) by (<%= str %>/<%= int %>/<%= con %>/<%= per %>/<%= attrs %>).
> Two stats, same bonus:
> Increases (Strength/Intelligence/Constitution) and (Intelligence/Constitution/Perception) by <%= attrs %> each.
> Two stats, different bonus
> Increases (Strength/Intelligence/Constitution) by (<%= str %>/<%= int %>/<%= con %>) and (Intelligence/Constitution/Perception) by (<%= int %>/<%= con %>/<%= per %>)].
> 
> Long form:
> Confers no benefit. / Increases (Strength/Intelligence/Constitution/Perception/all attributes) [and (Intelligence/Constitution/Perception)] by (<%= str %>/<%= int %>/<%= con %>/<%= per %>/<%= attrs %>) [each/and (Intelligence/Constitution/Perception) by (<%= int %>/<%= con %>/<%= per %>)].
> 
> .3. Event/Origin:
> (Month) (Year) Subscriber Item.
> Limited Edition (Year) (Season) Gear.


Alys:
I've tested on my local install and this compiles without errors. I haven't examined all the text changes in detail but the ones I have looked at look good. I like the consistency.

I'm going to merge this now so that @ShilohT doesn't have to deal again with more merge conflicts from any future changes that others might make. If there's any text changes that are considered inappropriate by anyone, we can adjust them later. Tier++. Thank you ShilohT!

@Lemoness, you might like to look at [the wording changes](https://github.com/HabitRPG/habitrpg-shared/pull/372/files) and ShilohT's questions above, and especially the detailed, thorough suggestions for a stylesheet to maintain consistency.

For reference, here's the guild conversation that ShilohT referred to:

```
Sulfirh:
Oh, talking about things which could be harmonized ! I realized some
time ago that a part of the Mystery Items do not have the note
"<Month>2014 Subscriber Item. Confers no benefit.". I finally
completed this on the French site, but if a Blacksmith want to make it
official by completing the original strings, that could be great.
(Even if I'm extremely sad to not have the first set, it's nice to see
this note on an old item and feel nostalgic ^^')

shiloh:
Would it be okay if I went through and standardized gear.json? Besides
the missing provenance notes that @Sulfirh mentioned, there are also a
lot of wording inconsistencies (For example: should special items be
called 'Gear', 'Outfit', 'Item', or 'Armor'? All four of these terms
are being used right now.) and formatting inconsistencies (The only
way someone has not written a note is with the Description last). I
would also throw a template comment at the bottom, so people will know
how to format in the future (Description; Effect; Event?).

Lemoness:
@Sulfirh Heh, did we sound like you'd expected? Also, re: Mystery
Items, I think that would be fine. For a while it was being added
automatically, but then it wasn't - kinda a crossed wire between me
and Tyler - so standardization is fine.
```

PR #377, opened 2014-12-04
Prettify Market Gem

SabreCat:
Adds a margin to the purchasable Gem in the Market so it doesn't look so wonky compared to other market items.

Before:
image

After:
image

I couldn't find anywhere else this particular zoom level of the Gem icon was used, but keep an eye out!

### DIFF
--- a/css/index.css
+++ b/css/index.css
@@ -40,3 +40,8 @@
 [class*="Mount_Head_"], [class*="Mount_Body_"]{
   margin-top:18px; /* Sprite accommodates 105x123 box */
 }
+
+.Pet_Currency_Gem {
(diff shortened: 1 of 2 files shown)


crookedneighbor:
:+1:

crookedneighbor:
@SabreCat while you're at it, I believe the beastmaster key could use the same margin treatment.

SabreCat:
@crookedneighbor: Ah! I hadn't noticed because I didn't have a full pet stable. Will check it out, thanks!

SabreCat:
Going to merge this as of the working commit, and futz with a fix to the pet key in a branch where I don't have an active PR.

PR #382, opened 2014-12-09
Betaveros cap crit multiplier PR into rebalancing-branch

crookedneighbor:


Alys:
It's gotten mixed up with other commits, which I'd like to avoid because there'll be a lot of PRs for this project and so I want each one to be a simple as it can be to avoid headaches. I'll rebase the feature branch soon and then you or I or betaveros can try again.

crookedneighbor:
@Alys Sure, that sounds good.

PR #386, opened 2014-12-11
Update communityGuidelines.json

lioman:
I think there is at least one word missing. (I' ve added two)

### DIFF
--- a/locales/en/communityGuidelines.json
+++ b/locales/en/communityGuidelines.json
@@ -76,7 +76,7 @@
     "commGuidePara044": "The following users are members of the HabitRPG repo:",
     "commGuideHeadingWiki": "Wiki",
     "commGuidePara045": "<strong>The HabitRPG wiki collects information about the site.</strong> It also hosts a few forums similar to the guilds on HabitRPG. Hence, all the Public Space rules apply.",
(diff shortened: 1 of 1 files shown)


Alys:
Hi @lioman!  "All things [noun]" is a phrase that means, roughly, "all things related to [that noun]". For example, see point 1.5 in http://www.oxforddictionaries.com/definition/english/thing but it might not be in common use.

@lemoness, this is a proposed change to the Community Guidelines, so we won't merge or close without your say-so.

lioman:
That is something that you can not do with a noun in German, that for I stumbled over the sentence while I try to translate it in transifex. And I did not know that you can do this in English. With HabbitRPG it sound a little bit "bumby". The example sentences under point 1.5 are clearer

Alys:
You can definitely change the wording as needed in translations!

"Bumby" is a word that I don't know. :) What does it mean?

lioman:
If you drive over a street with many holes in it - it is bumby

Alys:
Oh! I should have realised. It's spelled "bumpy" though. :) Your comment makes sense.

lioman:
That was a typo. I should not comment on language issues when I'm tired and should learn for exams. I push the "-"-button on HabitRPG

lemoness:
Maybe "all things related to HabitRPG"?

deilann:
I don't think that's necessary. It's pretty well understood by English
speakers, and translators will translate it appropriately in their language?

crookedneighbor:
I agree with @deilann, it's a common idiom in English.

PR #398, opened 2014-12-18
Added phantomjssmith to solve `Fatal error: Unsupported interlace method`

crookedneighbor:
Was getting `Fatal error: Unsupported interlace method` whenever I would run `grunt sprite`.

I was able to solve this by first running `npm install -g phantomjs`, then `npm install phantomjssmith --save`.

Devs who are compiling the sprites. Does that make sense to you? Should I add a line to the README about it?

### DIFF
--- a/package.json
+++ b/package.json
@@ -3,11 +3,12 @@
   "version": "0.0.0",
   "dependencies": {
     "browserify": "~3.30.2",
-    "coffeeify": "0.6.0",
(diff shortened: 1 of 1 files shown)


crookedneighbor:
I had a goal of closing some of these PRs in anticipation of removing habitrpg-shared, and only ended up making an additional one...

PR #406, opened 2014-12-23
update winter 2013 to winter 2014 to satisfy new naming conventions

joshbDev:
as referenced in this problem: https://github.com/HabitRPG/habitrpg-shared/issues/405. I've changed all the naming conventions from winter 2013 to winter 2014 to satisfy the new naming conventions.


crookedneighbor:
@SabreCat this satisfies the change, right? Nothing was missed?

Alys:
I'm very sorry @thehindenburg but we've decided to put both years in the user-visible text. For example, the latest equipment would say "Limited Edition 2014-2015 Winter Gear." If you'd like to make a PR for that change, we would welcome it!

PR #409, opened 2014-12-24
Winter 2013 --> Winter 2014

taldin:
Per habitrpg-shared #405


Alys:
I'm sorry @taldin but we've decided to put both years in the user-visible text. For example, the latest equipment would say "Limited Edition 2014-2015 Winter Gear."

It's possible that @thehindenburg will be making this change, since they had a competing PR at https://github.com/HabitRPG/habitrpg-shared/pull/406 :)

PR #412, opened 2014-12-25
Restored missing quote

gitte-k:
UUID _[edited by admin to remove wrong uuid]_


gitte-k:
Pasted wrong uuid, it's 3ecf07b0-4bea-4e6b-b0a0-4cb6e29a5699

Alys:
@lacrimae I checked your authentication data in the database and the first code you posted was your API token. Since that's like a password and needs to be kept secret, I have assigned a new token to you. If you were using your uuid and token in any third party apps, you'll need to get the new token from Settings -> API.

gitte-k:
Ah, thanks. Fail paste!

PR #418, opened 2014-12-29
Change valorousPresence to utilize diminishing returns. Keep results at ...

Verabird:
...200 str the same. (buffs lower levels, nerfs higher levels) References https://github.com/HabitRPG/habitrpg/issues/4418

### DIFF
--- a/script/content.coffee
+++ b/script/content.coffee
@@ -551,8 +551,9 @@ api.spells =
       notes: t('spellWarriorValorousPresenceNotes')
       cast: (user, target) ->
         _.each target, (member) ->
+          bonus = user._statsComputed.str
           member.stats.buffs.str ?= 0
-          member.stats.buffs.str += Math.ceil(user._statsComputed.str * .05)
+          member.stats.buffs.str += diminishingReturns(bonus, 20, 200)
 
     intimidate:
       text: t('spellWarriorIntimidateText')


Alys:
Thank you @Verabird, this does what we asked for! If we end up using it on the production site (which I currently think we probably will), this will count towards a contributor tier.

I'll add back Math.ceil around the diminishing returns function otherwise the Stats page shows buffs like 8.742459823401 :)

For those who might want to know:
At a STR of about 15, the old skill adds about 1, the new adds about 2.
At a STR of about 260, the old skill adds about 14, the new adds about 12.
At a STR of about 650, the old skill adds about 33, the new skill adds about 16. You could still get ridiculously high stats with this skill but I think you'd have to cast it twice as much to get up to 650, and a lot more than twice as much to get to the very high STRs that some people have reported.

Verabird:
Great! Knowing that this is what what needed (with Math.ceil) I can add
diminishing returns to the rest of the spells that haven't been done yet. I
will plan to keep scaling changes similar to this one unless otherwise
specified in the notes.

Alys:
Oh thank you so much!

Alys:
@Verabird For all of your changes for skills rebalancing and related code, you are now at Tier 5 and halfway towards Tier 6. Thank you very much for all of this!
([Change valorousPresence to utilize diminishing returns](https://github.com/HabitRPG/habitrpg-shared/pull/418), [Add diminishing Returns to several skills](https://github.com/HabitRPG/habitrpg-shared/pull/426), [Missed dailies decrease MP gained at cron](https://github.com/HabitRPG/habitrpg-shared/pull/448), [Adjust calculation of cronMP from checklists](https://github.com/HabitRPG/habitrpg-shared/pull/452), [Stealth allows checklists to reset at cron, etc](https://github.com/HabitRPG/habitrpg/pull/4736))

PR #419, opened 2014-12-30
Fix tests

crookedneighbor:
Corrected some failing tests.

Mostly the [user.ops.revive function requires that the user has 0 hp](https://github.com/HabitRPG/habitrpg-shared/blob/5e168c41cecc4626e9fd1914c4dffd86eb4d640c/script/index.coffee#L402) to work.

The other failing test expected for `quest.drop.items` to be an array, but the Basi list quest doesn't have an items property, so I just had it check to make sure `quest.drop.items` exists first.

Could not figure out [this one though](https://github.com/HabitRPG/habitrpg-shared/blob/develop/test/algos.mocha.coffee#L194). It's still failing inexplicably.


paglias:
:+1: :+1: i've fixed the remaining test here https://github.com/HabitRPG/habitrpg-shared/commit/0e266f917ef20f245a514dc25090e7a7905512c6

PR #422, opened 2014-12-30
Resize special items

crookedneighbor:
I resized the images for special items so they'd all be the same size as the gems.

screen shot 2014-12-30 at 12 04 32 pm

Spooky Sparkles and Snowballs have a separate image for when they appear in the rewards column, so the change does not affect them there.

The only other place one of the items appears is in the rebirth modals, which gives it just a hair more extra padding, which I like:
## Before

screen shot 2014-12-30 at 12 09 40 pm
## After

screen shot 2014-12-30 at 12 09 46 pm
## Before

screen shot 2014-12-30 at 12 22 24 pm
## After

screen shot 2014-12-30 at 12 15 54 pm


SabreCat:
:+1:

We should eventually redraw the Fortify Potion... it looks very strangely scaled compared to the other items!

crookedneighbor:
I think it looks ok actually. I kind of like it being tiny. For some reason, and I don't know why, it reminds me of the Felix Felicis potion in Harry Potter.

SabreCat:
It's not so much the size that bothers me as the fact that the border on the left side is wider than on the right, making it look misshapen.

crookedneighbor:
Oh, you're right. I can just fix that now, in this PR, if you'd like.

SabreCat:
Sure!

crookedneighbor:
Done: https://github.com/HabitRPG/habitrpg-shared/pull/422/files#diff-3

crookedneighbor:
Updated the pr to include the new years cards and valentine cards.

screen shot 2014-12-31 at 8 49 00 am

SabreCat:
:+1:

PR #424, opened 2014-12-30
Updated spritesmith

crookedneighbor:
Updated spritesmith the the latest version, adjusted the gruntfile to accommodate the breaking changes that v3 brought in, and specified the engine as phantomjssmith.

Now all the weird interlace and phantomjs crashing errors I was getting on `grunt sprite` are fixed.

### DIFF
--- a/Gruntfile.js
+++ b/Gruntfile.js
@@ -23,8 +23,9 @@ module.exports = function(grunt) {
     var sliced = images.slice(i * (images.length/COUNT), (i+1) * images.length/COUNT)
     sprite[''+i] = {
       src: sliced,
-      destImg: 'dist/spritesmith'+i+'.png',
-      destCSS: 'dist/spritesmith'+i+'.css',
+      dest: 'dist/spritesmith'+i+'.png',
+      destCss: 'dist/spritesmith'+i+'.css',
+      engine: 'phantomjssmith',
       algorithm: 'binary-tree',
       padding:1,
       cssTemplate: 'css/css.template.mustache',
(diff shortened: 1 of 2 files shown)


crookedneighbor:
Note: Anyone that will be running the `grunt sprite` command will need to update their node modules.

crookedneighbor:
@SabreCat, you're one of the more frequent users of the `grunt sprite` command, right? Is this cool with you?

SabreCat:
If all it takes is an `npm update` or similar, that's no problem.

crookedneighbor:
@SabreCat Did you want to check it out and make sure all is hunky dory in your system before I merge it in? This is one of the few things that I think differs from system to system, at least, I had some trouble with the old version of spritesmith. 

IT SHOULD BE FINE. But I'd prefer not to merge anything in that would hamper you when compiling sprites.

SabreCat:
Grabbed the commit, ran `npm update`, ran `grunt full` with a couple of new images in the tree: worked fine! Will merge.

paglias:
@crookedneighbor it's possible that for some reason this is preventing the site from being deployed. If I revert this PR do I need to run anything after that to make sure it works with the reverted code?

crookedneighbor:
You know, it might be.

I think a dependency of phantomjssmith is that phantomjs is installed on the system, so it might be freaking out if it can't find that. 

The only extra thing you'd need to do, if you updated the node modules, is to remove `node_modules/grunt-spritesmith`, and run `npm install` again.

PR #428, opened 2015-01-01
add remove-watch and use-timeout attribute to the markdown directive

negue:
Many angular watches are of the markdown directive, i've added two attributes which could improve step-by-step the performance.

For example if we use-timeout, we could delay the markdown parsing for some ms, and not start every markdown parsing instantly.

Also not every markdown needs to watch for changes, here comes the remove-watch attribute to help, for example this could be used on the chat messages 

Ping @paglias @crookedneighbor @lefnire  Any other ideas on this directive which I should implement? If not I'll merge this in 1-2 days :)

### DIFF
--- a/script/directives.js
+++ b/script/directives.js
@@ -92,7 +92,7 @@
         var removeWatch = !!scope.$eval(attrs.removeWatch);
         var useTimeout = !!scope.$eval(attrs.useTimeout);
 
-        var doRemoveWatch = scope.$watch(attrs.ngModel, function(value, oldValue) {
+        var doRemoveWatch = scope.$watch(attrs.text, function(value, oldValue) {
           var replaceMarkdown = function(){
 
             var markdown = value;
--- a/script/directives.js
+++ b/script/directives.js
@@ -85,26 +85,46 @@
     };
   }();
 
-  habitrpg.directive('markdown', function() {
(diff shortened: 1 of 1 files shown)


negue:
While optimizing the mobile code, I found that using ngModel we would still have one extra watch. which can't be removed (atleast with my knowledge^^). 

We would need to change the ng-model to something like "text" ?

paglias:
@negue why don't use one time bindings?

negue:
Even with `<markdown ng-model="::var" />` we still have the ng-model watch and the inner $watch which points also to scope.ngModel

paglias:
@negue :+1:

script/directives.js
```diff
@@ -85,26 +85,46 @@
...
+            var html = md.toHtml(markdown);
+
+            html = html.replace(userHighlight, "<u>@"+userName+"</u>");
+
+            html = html.replace(' href',' target="'+linktarget+'" href');
+            element.html(html);
+
+            if(removeWatch)
+            {
+              doRemoveWatch();
```
lefnire, line 111: Not sure I understand this part. If `removeWatch` then call `doRemoveWatch()` recursively?

negue, line 111: once the markdown is parsed and replaced to that element, then look if this directive shouldn't longer look for changes and then call that $watch destructor

PR #429, opened 2015-01-01
Issue4454habitrpg

melissamcewen:
I added the immediate use variable

### DIFF
--- a/dist/habitrpg-shared.js
+++ b/dist/habitrpg-shared.js
@@ -12436,7 +12436,6 @@ api.spells = {
       text: t('spellSpecialSaltText'),
       mana: 0,
       value: 5,
-      immediateUse: true,
       target: 'self',
       notes: t('spellSpecialSaltNotes'),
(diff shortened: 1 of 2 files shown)


crookedneighbor:
Awesome. Can you add it to the opaque potion and nye card spells as well?

melissamcewen:
@crookedneighbor oh yeah, those haha. Added.

melissamcewen:
Would it also make sense to add info about it costing coins in the description?

crookedneighbor:
I don't think so. I think it'd be redundant to include it, because the only places you see it are in the rewards column, where it's very obvious that it costs 5 gold:

screen shot 2015-01-01 at 4 30 18 pm

And the seasonal shop for New Year Cards:

screen shot 2015-01-01 at 4 30 29 pm

melissamcewen:
ah mine didn't have the coin but I then realized my shared repo was out of date

PR #434, opened 2015-01-11
change G to Gem to remove confusion with Gold

Alys:
In the Subscription page's pop-up for "Buy Gems with Gold", I've changed "G" to "Gems" to avoid confusion with "Gold" (one user was asking about that in Newbies).

screen shot 2015-01-11 at 10 57 59 am


deilann:
+1

This came up once before.

PR #438, opened 2015-01-14
Prevent type changing

crookedneighbor:
Brought in @lefnire's change to prevent task type from being changed on a task.


crookedneighbor:
Closes https://github.com/HabitRPG/habitrpg/issues/4514

PR #442, opened 2015-01-16
Triad bingo

crookedneighbor:
Companion for: https://github.com/HabitRPG/habitrpg/pull/4529


script/index.coffee
```diff
@@ -774,6 +776,10 @@ api.wrap = (user, main=true) ->
           if not user.achievements.mountMasterCount
             user.achievements.mountMasterCount = 0
           user.achievements.mountMasterCount++
+          if giveTriadBingo
```
crookedneighbor, line 779: @deilann is this acceptable to you? 
- Set `giveTriadBingo = true`
- for every pet, check if `pet == -1`, and if so set `giveTriadBingo = false`. 
- if `giveTriadBingo == true` at the end, then increase count of triadBingoCount

You can't use `releaseBoth()` without having all 90 mounts, so there's no need to check for that.

crookedneighbor:
When to add in tests for triadBingo, but it turns out that none of the tests for beastMaster and mountMaster actually work... `expect(user.achievements.beastMaster).to.be.ok` and `expect(user.achievements.beastMaster).to.not.be.ok` both evaluate true for some reason...

PR #445, opened 2015-01-18
fix typo in README.md

j-wilkins:
in the banner about migrating this repo migrating was typo'd as migraring.

:purple_heart:


crookedneighbor:
Haha, thanks for catching that. 

Won't be enough to merit a contributor tier, but if you post your user id here, I'll definitely notate it!

PR #426, opened 2014-12-31
Add diminishing Returns to several skills

Verabird:
References https://github.com/HabitRPG/habitrpg/issues/4419, https://github.com/HabitRPG/habitrpg/issues/4417, https://github.com/HabitRPG/habitrpg/issues/4416, https://github.com/HabitRPG/habitrpg/issues/4415, https://github.com/HabitRPG/habitrpg/issues/4414, https://github.com/HabitRPG/habitrpg/issues/4412.

I tried to keep values in line with what was there previously or in the notes. However, the diminishingReturns function generally buffs at low levels and nerfs at high levels compared to a linear function.

PR #448, opened 2015-01-22
Missed dailies decrease MP gained at cron

Verabird:
This is in reference to https://github.com/HabitRPG/habitrpg/issues/4473.

Undone dailies decrease MP gained at cron by counting checked dailies and due but unchecked dailies and multiplying the MP gain by checked/(checked+dueUnchecked). Checklists are included by subtracting fractions of due but unchecked. Gray dailies which are checked off will add to the count of checked, but are otherwise ignored. (Also I left in comments that were notes to myself figuring out what was already there)

I did some tests, please let me know if there are others I should run.
## Test Results follow

Tests on Thu:
Run cron by "+1 missed day" button
Max MP < 100
gray daily means deactivated for Wed.

Setup:
1 regular gray daily
1 gray daily with checklist of 3 items
1 regular active daily
1 active daily with checklist of 4 items

Tests:
    check off both active, but not checklists:Gain 10 MP, DueUnchecked=0, Checked=2
    check off 1 of active checklist, and the regular active daily: Gain 5.7.. MP, DueUnchecked=0.75, Checked =1
    check off 3 of active checklist and the regular active daily: Gain 8 MP, DueUnchecked=0.25, Checked=1
    check off regular active and entire active checklist but not main: Gain 10 MP, DueUnchecked=0, Checked = 1
    check off regular active daily, none of checklist (inc. main): Gain 5 MP, DueUnchecked=1, Checked=1
    check off Nothing: Gain 0 MP, DueUnchecked=2, Checked=0
    Check off 2 on gray checklist, and both mains for active: Gain 10 MP, DueUnchecked=0, Checked=2
    Check off 2 on gray checklist, regular active, and nothing else: Gain 5 MP, DueUnchecked=1, Checked=1
    check off both gray main, regular active, nothing else: Gain 7.5 MP, DueUnchecked=1, Checked=3
    check off gray regular, regular active, nothing else: Gain 6.66... MP, DueUnchecked=1, Checked=2

Setup:
Only Gray Dailies
Test:
    Nothing checked: Gain 10 MP, DueUnchecked=0, Checked=1

### DIFF
--- a/script/index.coffee
+++ b/script/index.coffee
@@ -1071,6 +1071,7 @@ api.wrap = (user, main=true) ->
           user.stats.mp = user._statsComputed.maxMP if user.stats.mp >= user._statsComputed.maxMP
           user.stats.mp = 0 if user.stats.mp < 0
 
+        # ===== starting to actually do stuff, most of above was definitions ==========
         switch task.type
           when 'habit'
             changeTaskValue()
@@ -1465,8 +1466,12 @@ api.wrap = (user, main=true) ->
               thatDay = moment(now).subtract({days: n + 1})
               scheduleMisses++ if api.shouldDo(thatDay, repeat, user.preferences)
           if scheduleMisses > 0
-            perfect = false if type is 'daily'
-            dailyDueUnchecked++ if type is 'daily'
+            if type is 'daily'
+              perfect = false 
+              if task.checklist?.length > 0  # Partially completed checklists dock less mana points
+                dailyDueUnchecked += (1 - _.reduce(task.checklist,((m,i)->m+(if i.completed then 1 else 0)),0) / task.checklist.length)
+              else
+               dailyDueUnchecked += 1
             delta = user.ops.score({params:{id:task.id, direction:'down'}, query:{times:scheduleMisses, cron:true}});
             user.party.quest.progress.down += delta if type is 'daily'
         else
@@ -1481,9 +1486,6 @@ api.wrap = (user, main=true) ->
           #get updated value
             absVal = if (completed) then Math.abs(task.value) else task.value
             todoTally += absVal
-
-      console.log("DueUnchecked: " + dailyDueUnchecked)
-      console.log("Checked: " + dailyChecked)
 
       user.habits.forEach (task) -> # slowly reset 'onlies' value to 0
         if task.up is false or task.down is false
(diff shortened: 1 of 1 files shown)


Alys:
I have noted this as giving credit to a contributor tier if we use it (which is likely). Still testing it locally.

script/index.coffee
```diff
@@ -1463,9 +1466,16 @@ api.wrap = (user, main=true) ->
...
+            if type is 'daily'
+              perfect = false 
+              if task.checklist?.length > 0  # Partially completed checklists dock less mana points
+                dailyDueUnchecked += (1 - _.reduce(task.checklist,((m,i)->m+(if i.completed then 1 else 0)),0) / task.checklist.length)
+              else
+               dailyDueUnchecked += 1
             delta = user.ops.score({params:{id:task.id, direction:'down'}, query:{times:scheduleMisses, cron:true}});
             user.party.quest.progress.down += delta if type is 'daily'
+        else
+          dailyChecked++
```
Alys, line 1478: @Verabird Is dailyChecked incremented by completed To-Dos? I _think_ it is from reading the code, but I can't test because I've broken my local install (fourth time in three days).

If you need to change the code for this, I think it's worth also reversing the `unless completed .... else ....` statement to make it if-else. Now that we have an else branch, there's no pressing need to use `unless`, and the logic will be easier to follow if it's an if-else statement (especially since we have several programmers who aren't very familiar with coffeescript).

Verabird, line 1478: You're right. Thanks for catching this Alys. I'll fix it and change to an
if-else statement later today.

Alys:
@Verabird I've done some testing and most of this PR is what we want but there's some odd behaviour with checklists under some circumstances.

When checklist items are ticked, they **_do**_ reduce MP loss if another main Daily is ticked (dailyDueUnchecked and dailyChecked are both non-zero), but they **_don't**_ reduce MP loss if no main Dailies are ticked (dailyDueUnchecked is non-zero but dailyChecked is zero). This seems to be a side-effect of this code: `user.stats.mp += ... * dailyChecked / (dailyDueUnchecked + dailyChecked)` -- when dailyChecked is zero, dailyDueUnchecked can't have any effect.

Also, the effect of the checklist items is different for health compared to MP. I'm probably going to explain this really badly, so do tell me if it's unclear! 

I'm using the term "cron-MP" to describe the amount of MP that the user gains on cron.

I think what we want is for checklists to affect the reduction in cron-MP in exactly the same way that they affect the reduction in Health-loss. Some examples of how I think it should work, compared to how it does work currently:
1. If you have ticked off all the checklist items but not the Daily itself, you lose no Health for that Daily and you lose no cron-MP (this does happen with your code and is good).
2. If you have ticked off half the checklist items, you lose half as much Health as you would if you had ticked none of them (this does happen), and you should lose half as much cron-MP (this does not happen; you lose a different proportion of cron-MP, depending on exactly how many checklist items there are).
3. If you have four checklist items and you tick off three of them, you lose one-quarter as much Health as you would if you had ticked none of them (this does happen), and you should lose one-quarter as much cron-MP (this does not happen; you lose two-fifths as much cron-MP - it's as if the Daily itself is being counted as a checklist item when the proportion of cron-MP is being calculated).

I hope that makes sense... :)

Verabird:
@Alys Yes, this does make sense. I think having checklists partially increment dailyChecked will fix the problem. I'll run these tests before doing another PR to be sure, though.

Alys:
Tier 5 awarded and a note has been made that Verabird is halfway towards Tier 6 for these changes:
[Change valorousPresence to utilize diminishing returns](https://github.com/HabitRPG/habitrpg-shared/pull/418)
[Add diminishing Returns to several skills](https://github.com/HabitRPG/habitrpg-shared/pull/426)
[Missed dailies decrease MP gained at cron](https://github.com/HabitRPG/habitrpg-shared/pull/448)
[Adjust calculation of cronMP from checklists](https://github.com/HabitRPG/habitrpg-shared/pull/452)
[Stealth allows checklists to reset at cron, etc](https://github.com/HabitRPG/habitrpg/pull/4736)

PR #454, opened 2015-01-27
updating rebirth to account for max levels 100+

deilann:
Rebirth stacks whenever you hit a higher level than your previous rebirth _or_ you hit a level greater than or equal to 100 _and_ records your highest level correctly

This PR also includes some language tweaking to make Rebirth stuff make more sense with that.


deilann:
piing @SabreCat

script/index.coffee
```diff
@@ -516,13 +513,14 @@ api.wrap = (user, main=true) ->
...
-        # Award an achievement if this is their first Rebirth, or if they made it further than last time
+        # Award an achievement if this is their first Rebirth, or if they made it further than last time, or if they hit over 100
         if not (user.achievements.rebirths)
           user.achievements.rebirths = 1
           user.achievements.rebirthLevel = lvl
-        else if (lvl > user.achievements.rebirthLevel or lvl is 100)
-          user.achievements.rebirths++
+        else if (lvl > user.achievements.rebirthLevel)
           user.achievements.rebirthLevel = lvl
+        if (lvl >= 100)
```
SabreCat, line 522: Way I'm reading this, if the user buys an Orb for the first time at level 100+, they will get 2 Rebirth achievements. The `if not` clause sets it to 1, then the `if (lvl >= 100)` bit increments it by one.

I would instead change `else if (lvl > user.achievements.rebirthLevel or lvl is 100)` to `else if (lvl > user.achievements.rebirthLevel or lvl >= 100)`.

script/index.coffee
```diff
@@ -475,10 +475,7 @@ api.wrap = (user, main=true) ->
         if user.stats.lvl < 100
           user.balance -= 2
         # Save off user's level, for calculating achievement eligibility later
-        if user.stats.lvl < 100
-          lvl = user.stats.lvl
-        else
-          lvl = 100
+        lvl = user.stats.lvl
```
SabreCat, line 478: Because you're no longer capping this at 100, this bit isn't strictly necessary; you'd then use `user.stats.lvl` later instead of `lvl`.

SabreCat:
Cool! The first comment is a bug, the second one just a possibility/suggestion.

script/index.coffee
deilann, line 522: looks like there's actually another problem with it, too...

rebirths under 100 aren't going to stack.

SabreCat, line 522: Ah, indeed. I think my suggested fix would catch that too, though.

crookedneighbor:
@deilann About to open my PR to consolidate shared into main.

I've merged your PR into this branch: https://github.com/HabitRPG/habitrpg/tree/deilann-orb-rebirth-refactor

Going to close this for now, will assist if you need help refactoring

deilann:
thank you for doing this. i've just been dealing with some serious issues that have taken away my time to do pretty much _anything_. I felt bad about having a shared PR open.

crookedneighbor:
@deilann No worries dude. Take care of yourself.

PR #455, opened 2015-01-28
New tasks for onboarding

crookedneighbor:
ping @lemoness @Alys @sugarfiend @veeeeeee

I was wrong, the notes field is required for the default tasks, so I just created a temporary empty string for them until notes can be written.

Should there be any variation in value (and therefore color) of the tasks? Join the newbies guild inherited the value of -3 from the call mom task.

Anything else you want to change?

screen shot 2015-01-27 at 9 54 50 pm
screen shot 2015-01-27 at 9 54 57 pm

### DIFF
--- a/dist/habitrpg-shared.js
+++ b/dist/habitrpg-shared.js
@@ -14303,13 +14303,7 @@ api.userDefaults = {
     {
       type: 'todo',
       text: t('defaultTodo1Text'),
-      notes: t('defaultTodo1Notes'),
-      completed: false,
-      attribute: 'int'
-    }, {
-      type: 'todo',
-      text: t('defaultTodo2Text'),
-      notes: t('defaultTodo2Notes'),
+      notes: t('defaultTodoBlank'),
       completed: false,
       attribute: 'int'
     }, {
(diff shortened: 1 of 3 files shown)


crookedneighbor:
Incorporate some emoji?

Alys:
Oh heck! Something I meant to say yesterday: Do we really want people introducing themselves in the Newbies guild? I worry that that will lead to a lot of welcoming conversations and daily greetings, which is nice for socialness, but the advantage of the Newbies guild is that there's almost none of that so that answers remain visible for a long time.

Perhaps "Visit the Tavern and introduce yourself" and also "Join the Newbies guild when you have questions"?

Sorry this is so late. :(  Ignore me if no one agrees with me. :)

Alys:
+1 for learn emoji. People love it.

crookedneighbor:
What if we changed learn markdown to "Learn Markdown and use Emoji"?

lemoness:
1. Newbies guild should be last on the list (trying to rank them in
   importance). I am okay with it being just "Introduce yourself in the
   Tavern." We can always direct people to Newbies Guild if they have
   questions.
2. I would like us to get rid of the "learn markdown and use emoji" because
   we already have a lot of to-dos and I'm not convinced it's going to make
   Habit "stickier" - but that being said, I think having the sample tasks
   INCLUDE emoji in their names is a good idea!
3. Do you need sample notes before you can merge this, or is it okay?

deilann:
Any tasks that involve introducing oneself should involve a subtask of reading the community guidlines.

lemoness:
Maybe we should instead just make it "Join the Newbies Guild" so that they
have the option to ask questions if they need to, but we don't get flooded
with introductions anywhere.

lemoness:
(More than usual, obviously.)

AnitaYCheng:
(sugarfiend here - sorry I used a proper name for GitHub :P )

I agree in leaving "Join Newbies Guild/Tavern" for last. Learning how to use Habit at all should be priority.

Also, we could say "Start or join a Party" if they already have friends there. 

Meh on Markdown, yay on Emojis haha. (But that was just my personal onboarding process..)

Alys:
Not something we'd want to do immediately, but perhaps consider for later:

Ryan was talking a while back about having challenges for new users, the idea being that you join the challenge in your first month of using the site to learn more. We could put the more advanced things (markdown? parties? avatar customisation? read the wiki?) in challenge tasks, and have "join the ... challenge" as a default To-Do.

AnitaYCheng:
I like that idea @Alys! So it isn't like, something everyone has to go through to use HabitRPG in at most basic, but if you want to become a power user, it would give you an incentive to do it.

veeeeeee:
Let's try out parties as a first-step task, rather than something to do later by joining a challenge? We have a hunch that being in a party helps with retention.

AnitaYCheng:
But starting a party wouldn't necessarily fix that, since you'd have to find friends for your party. And how easy is it to find people from your Facebook friend list on Habit?

If anything, I would suggest joining a guild for a subject they're interested in and introducing themselves there. At least they'd already have something in common, instead of everyone being a Newbie and nobody really paying attention when half of them are just "Hi I'm new!". Also, it isn't very intuitive that there's messages waiting for you - I hope a notification alert (number of unread messages over the nav, like Facebook and Twitter do it) is in the works!

veeeeeee:
I'm really not certain that introducing themselves in a guild helps with retention, just from what I've seen. There are definitely personalities who stick around, but I think they tend to stick around generally in the social sector. (In other words, retention might be correlated with, but in most cases probably not caused by, guild participation)

On ease of finding people: not via Facebook, but you can extend an email invitation. When they join, you're all set to go with at least one Basi-list quest, so you can try out a quest which will also motivate everyone participating to log in every day.

AnitaYCheng:
If it isn't already implemented, there should be a checklist under "Start a party" that says "Invite friends," haha.

crookedneighbor:
@Alys, I love the idea of a challenge for newbies. I was going to suggest putting it in the Newbies guild, but there's already quite a few challenges in there. Also, we need to figure out how to scroll down to the correct challenge when it's clicked on or a newbie will never find it.

@lemoness or @veeeeeee, could you make a checklist of changes you want made? You can make a checklist like this:

```
- [ ] Checklist item 1
- [ ] Checklist item 2
- [ ] Checklist item 3
```

And that will turn into: 
- [ ] Checklist item 1
- [ ] Checklist item 2
- [ ] Checklist item 3

lemoness:
- [x] Remove "Join Newbies Guild"
- [x] Remove "Learn Markdown"
- [x] De-capitalize "Your Friends"

Some comments:
I like the idea of a Challenge that holds the more advanced tasks. Let's keep these initial To-Dos to the most basic, then - setting up the three task types and joining a party. (From the party screen, it gives you options to invite your friends, so it is straightforward; I don't think we need a checklist.) Five is a good number for Sample To-Dos.

Later on, we can finalize the Newbies Challenge and experiment with having that be one of these sample To-Dos - although by then we may have totally rehauled the tutorial.

lemoness:
@crookedneighbor Vicky says she has no changes, so go ahead and make the
changes/merge/push!

Alys:
@crookedneighbor If a challenge is created by a mod/staff and marked as an official challenge, I believe it sorts to the top of the list, which would make it easier to find.

PR #457, opened 2015-01-29
open external links

negue:
Fix for HabitRPG/habitrpg-mobile#338

Partial fix part 1 for HabitRPG/habitrpg-mobile#227  (only the stats page link left)

### DIFF
--- a/script/directives.js
+++ b/script/directives.js
@@ -85,7 +85,7 @@
     };
   }();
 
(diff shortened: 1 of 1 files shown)


paglias:
@negue partial because it only applies to markdown links?

negue:
@paglias yep, I'm currently changing / testing the stats page links using a jade mixin hope it'll work

PR #459, opened 2015-01-29
add translation strings

ghost:
pulled some text from jade templates and added to en translations
habitrpg issue #4522


crookedneighbor:
Merged.

PR #460, opened 2015-01-30
add a locale var to all translations

negue:


script/i18n.coffee
```diff
@@ -15,10 +15,13 @@ module.exports =
 
     locale = 'en' if (!locale? or (!module.exports.strings and !module.exports.translations[locale]))
     string = if (!module.exports.strings) then module.exports.translations[locale][stringName] else module.exports.strings[stringName]
+    
+    clonedVars = _.clone(vars) or {};
```
paglias, line 19: it's necessary to clone here or we could do in habitrpg-shared since it seems to be only needed for drops' notes?

negue, line 19: hm what do you mean? we should I do it then?

I thought If I not cloned it, then we could also get cross-request of different languages

